PR #75, opened 2023-04-20
change selector from `strimzi` to `rhacs`

stehessel:
Purge Kafka names take 2. After `rhacs` Grafana and Prometheus selectors in https://github.com/stackrox/acs-fleet-manager/pull/885 have now been deployed, we should be able to finally remove `strimzi`.


dhaus67 (approved):
No more `strimzi`
POGGIES

PR #76, opened 2023-04-21
ROX-10715: RHACS Central SLI, SLO alerts

0x656b694d:
RHACS Central error budget exhaustion alerts to track SLO violations.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -383,7 +383,7 @@ spec:
         - alert: Central availability error budget exhaustion - 90%
           annotations:
             message: "High availability error budget exhaustion for central. Current exhaustion: {{ $value | humanizePercentage }}."
-            sop_url: "https://github.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-018-rhacs-central-slo-alerts.md"
+            sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-018-rhacs-central-slo-alerts.md"
           expr: |
             central:slo:availability:error_budget_exhaustion >= 0.9
           labels:
@@ -400,7 +400,7 @@ spec:
         - alert: Central availability error budget exhaustion - 70%
           annotations:
             message: "High availability error budget exhaustion for central. Current exhaustion: {{ $value | humanizePercentage }}."
-            sop_url: "https://github.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-018-rhacs-central-slo-alerts.md"
+            sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-018-rhacs-central-slo-alerts.md"
           expr: |
             central:slo:availability:error_budget_exhaustion >= 0.7
           labels:
@@ -416,7 +416,7 @@ spec:
         - alert: Central availability error budget exhaustion - 50%
           annotations:
             message: "High availability error budget exhaustion for central. Current exhaustion: {{ $value | humanizePercentage }}."
-            sop_url: "https://github.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-018-rhacs-central-slo-alerts.md"
+            sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-018-rhacs-central-slo-alerts.md"
           expr: |
             central:slo:availability:error_budget_exhaustion >= 0.5
           labels:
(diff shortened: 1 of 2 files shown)


resources/prometheus/prometheus-rules.yaml
```diff
@@ -291,3 +291,139 @@ spec:
               The DB instance `{{ $labels.dimension_DBInstanceIdentifier }}` has scaled up as high as it can.
               Consider increasing the maximum ACU setting for the cluster.
             sop_url: "https://github.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-014-aws-rds-acu-utilization.md"
+
+    - name: rhacs-central.sli
```
stehessel, line 295: It would be nice have the `rhacs-*` labels on the recording rules. In particular, `rhacs_instance_id`, `rhacs_org_name`, `rhacs_org_id`, `rhacs_cluster_name`. This probably requires some join

resources/prometheus/unit_tests/RHACSCentralSLISLO.yaml
```diff
@@ -0,0 +1,143 @@
...
+        exp_alerts:
+          - exp_labels:
+              alertname: Central availability error budget exhaustion - 90%
+              service: central
+              severity: critical
+              team: appsre
+              namespace: rhacs-test1
+            exp_annotations:
+              message: "High availability error budget exhaustion for central. Current exhaustion: 100.2%."
+              dashboard: "https://grafana.app-sre.devshift.net/d/T2kek3H9a/central-slos?orgId=1&from=now-28d&to=now"
```
stehessel, line 44: These are the dashboard/runbooks for the acs-fleet-manager. I'd suggest removing the dashboard link, and creating an entry in https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks for the alert runbook. Also rename `runbook` to `sop_url` for consistency with the other data plane alerts.

resources/prometheus/prometheus-rules.yaml
0x656b694d: What about `rhacs_environment`? Does it make sense to include it here?

stehessel: Yes that one as well.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -291,3 +291,159 @@ spec:
...
+        - alert: Central availability error budget exhaustion - 90%
+          annotations:
+            message: "High availability error budget exhaustion for central. Current exhaustion: {{ $value | humanizePercentage }}."
+            sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-018-rhacs-central-slo-alerts.md"
+          expr: |
+            central:slo:availability:error_budget_exhaustion >= 0.9
+          labels:
+            service: central
+            severity: critical
+            team: appsre
```
stehessel, line 392: You can remove the `team` label. It's needed for some process in app-interface, but we don't use it for data plane alerts.

stehessel:
Imo the PR is looking good. Do you think it makes sense to loop in one of the SRE guys for review as well?

resources/prometheus/prometheus-rules.yaml
```diff
@@ -291,3 +291,157 @@ spec:
...
+          record: central:sli:availability:avg_over_time1h
+
+        - expr: |
+            avg_over_time(central:sli:availability[28d])
+          record: central:sli:availability:avg_over_time28d
+
+    - name: rhacs-central.slo
+      rules:
+        # Availability SLO
+        - expr: "0.99"
```
stehessel, line 363: Ah one more issue: The SLO target for Central is `99.95%`, not `99%` as for fleet-manager. So we need to adapt this number here, and then fix the numbers in the tests as well.

stehessel: Another option would be to keep the `99%` for now for Central as well, with the intention to increase the target gradually to `99.95%`. Reason being that I don't think `99.95%` is really realistic with the current Central architecture. What is your opinion on this?

0x656b694d: > The SLO target for Central is `99.95%`
Let's set it so and be reminded.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -291,3 +291,157 @@ spec:
...
               Consider increasing the maximum ACU setting for the cluster.
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-014-aws-rds-acu-utilization.md"
+
+    - name: rhacs-central.sli
+      rules:
+        # Pod ready SLI
+        # Central pod must be ready to count as available.
+        # This is a time series of 0s (down) and 1s (up).
+        - expr: |
+            clamp_max(sum by(namespace) (kube_pod_container_status_ready{container="central", pod=~"central-.*", namespace=~"rhacs-.*"}), 1) or vector(0)
```
stehessel, line 301:
```suggestion
            clamp_max(sum by(namespace) (kube_pod_container_status_ready{container="central", pod=~"central-.*", namespace=~"rhacs-.*"}), 1)
```

resources/prometheus/prometheus-rules.yaml
0x656b694d: No, let's keep 99%, as 99.95% seems unrealistic today.

stehessel (approved):
Can you rebase to master to fix the tests?

SimonBaeumer:
@0x656b694d You can dismiss my feedback, it is sufficient to me that @stehessel has approved it.

PR #82, opened 2023-04-28
ROX-16805: Fix Kubernetes mixins

stehessel:
A few fixes to the K8s mixin dashboards:

* OpenShift already provides Kubernetes mixin rules by default. To avoid duplication of metrics via recording rules, we should make sure not to generate the same recording rules here. Otherwise, `sum` and other aggregations are miscalculated.
* Set the grafana rate interval to `5m`. Not sure why this is needed, but Grafana's automagical `$__rate_interval` does not seem to work in all cases.
* Set the correct `containerfsSelector` for OpenShift clusters.
* `make generate` to update the dashboards. There were some bugs with wrong escapes that are fixed in the new mixin version.

Relevant files are `resources/index.json`, `resources/mixins/kubernetes/mixin.libsonnet`, and `resources/mixins/kubernetes/Makefile`.


resources/mixins/kubernetes/README.md
```diff
@@ -10,3 +10,8 @@ The Kubernetes mixin requires jsonnet to render its output resources:
 ## Update
 
 Execute `make generate` to update the mixin and re-generate all resources.
+
+## OpenShift platform metrics federation
+
+OpenShift already provides Kubernetes mixin rules by default. To avoid duplication of metrics via recording rules,
+no recording rules generated.
```
dhaus67, line 17: super nit:
```suggestion
no recording rules are generated.
```

dhaus67 (approved):

PR #85, opened 2023-05-04
ROX-16887: ignore first 5m of pod life

0x656b694d:
Ignore first 5 minutes of the centrals pod ready metric so that we don't count the startup non-ready time in SLO.

Correction after #76.


0x656b694d:
Closed in favor of #87.

PR #88, opened 2023-05-05
make acu alert more resilient against spikes

stehessel:
* Remove the `avg_over_time` because AWS already aggregates metrics over a five minute interval.
* Switch the alert from `_maximum` to `_average` to be more resilient against temporary ACU spikes. 
* Add a filter based on the instance DB label. This is currently redundant, but just in case we ever export other ACU metrics in the future - for example total cluster ACU.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -281,8 +281,8 @@ spec:
       rules:
         - alert: AWSRDSACUUtilization
           expr: |
-            avg_over_time(aws_rds_acuutilization_maximum[5m]) > 95
-          for: 1h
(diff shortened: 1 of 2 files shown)


resources/prometheus/prometheus-rules.yaml
```diff
@@ -281,8 +281,8 @@ spec:
       rules:
         - alert: AWSRDSACUUtilization
           expr: |
-            avg_over_time(aws_rds_acuutilization_maximum[5m]) > 95
-          for: 1h
+            aws_rds_acuutilization_average{dimension_DBInstanceIdentifier=~"rhacs-.*"} > 95
+          for: 15m
           labels:
             severity: critical
```
vladbologa, line 287: Not related to the change, but is this error really critical?

vladbologa (approved):

PR #89, opened 2023-05-05
use deployment status instead of pod status

stehessel:
Use `kube_deployment_status_replicas_ready` to more accurately track the SLI in cases the pod status is not defined - for example deployment upgrades. Note that in most cases both metrics yield identical results.


ivan-degtiarenko (approved):
LGTM! Quick question - when wouldn't they yield identical results considering we have 1 replica?

stehessel:
Michael pointed out that there can be a difference in some cases. I think one example would be if the deployment scales down the pods, then there is no pod status metric, but the deployment status metric would show zero.

PR #91, opened 2023-05-08
ROX-16884 Add alerts when cpu/memory commitment per availability zone…

ludydoo:
… exceeds thresholds

### DIFF
--- a/resources/mixins/kubernetes/jsonnetfile.lock.json
+++ b/resources/mixins/kubernetes/jsonnetfile.lock.json
@@ -18,7 +18,7 @@
           "subdir": "grafana-builder"
         }
       },
-      "version": "1cb605505efe5a630f00993ff611a241fdcc264f",
+      "version": "d303b2031264728728dd1e1c05f74f67027139f6",
       "sum": "tDR6yT2GVfw0wTU12iZH+m01HrbIr6g/xN+/8nzNkU0="
     },
     {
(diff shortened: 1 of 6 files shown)


resources/mixins/kubernetes/jsonnetfile.lock.json
```diff
@@ -18,7 +18,7 @@
           "subdir": "grafana-builder"
         }
       },
-      "version": "1cb605505efe5a630f00993ff611a241fdcc264f",
+      "version": "d303b2031264728728dd1e1c05f74f67027139f6",
```
stehessel, line 21: Why did you update the mixin?

resources/prometheus/prometheus-rules.yaml
```diff
@@ -476,3 +476,141 @@ spec:
             rhacs_org_id: "{{ $labels.rhacs_org_id }}"
             rhacs_cluster_name: "{{ $labels.rhacs_cluster_name }}"
             rhacs_environment: "{{ $labels.rhacs_environment }}"
+    - name: az-resources
+      rules:
+        - alert: WorkerNodesByAZMemoryQuotaOverCommit
+          expr: |
+            (
```
stehessel, line 483: I think these promQL queries a bit too lengthy. Could you try to break them down with recording rules?

stehessel:
Can you add unit tests for the new alerts?

stehessel:
Also I'd recommend to set up your editor to clip unnecessary whitespaces ;-)

resources/prometheus/prometheus-rules.yaml
stehessel: Maybe also keep in mind that we'd like to use some of these queries as part of our dashboard, so the shorter the better 😅

resources/mixins/kubernetes/jsonnetfile.lock.json
ludydoo: I just ran `make` and it did that all by itself, perhaps a new version?

stehessel (approved):
LGTM 🚀

PR #92, opened 2023-05-09
Reduce RHACSFleetshardSyncReconciliationErrors sensitivity

kylape:
Change threshold from 1% error rate to 10%.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -148,7 +148,7 @@ spec:
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-005-fleetshard-sync-unavailable.md"
         - alert: RHACSFleetshardSyncReconciliationErrors
           expr: |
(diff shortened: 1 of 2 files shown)


stehessel (approved):

kylape:
My understanding of `27% over the last 10 minutes`:

These are the numbers in the time series generated in the unit test by the expression `0+0x5 0+0.03x15`:

0
0
0
0
0
0
0.03
0.06
0.09
0.12
0.15
0.18
0.21
0.24
0.27

Each reading represents one minute.  The specific test for this alert take its reading at minute 15, which calculates the error ratio over the last 10 minutes, which is 0.1125, just over the threshold of 0.1.  The message in the alert, though takes the most recent value of the time series, which is 0.27 (`27%` after using the `humanizePercentage` function).

stehessel:
In this case each value (like `0.27`) already represents the last 10 minutes because of the `rate(...[10m])` that goes into the metric definition. Because of how Prometheus works, every minute a new value appears, which is calculated based on the previous 10 minutes. So there is some overlap in the calculation of each point.

PR #95, opened 2023-05-17
ROX-17093: only count server error codes in grpc SLI

stehessel:
We don't want to count client errors (4xx equivalents) into the error rate SLI. Furthermore, this change allows us to modify the behavior of Central when the Scanner semaphore limit is hit. In this case we currently return `grpc_code=Unavailable`, but we are thinking about changing the return code to `grpc_code=ResourceExhausted`. Expected retries due to the semaphore would then no longer count as unavailable.

See https://grpc.github.io/grpc/core/md_doc_statuscodes.html for a list of all grpc codes and their meaning.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -307,8 +307,8 @@ spec:
         # GRPC
         - expr: |
             sum by (namespace, rhacs_instance_id, rhacs_org_id, rhacs_org_name, rhacs_cluster_name, rhacs_environment)
-            (rate(grpc_server_handled_total{namespace=~"rhacs-.*", job="central", grpc_type="unary", grpc_code="OK"}[10m]))
-          record: central:grpc_server_handled:OK:rate10m
+            (rate(grpc_server_handled_total{namespace=~"rhacs-.*", job="central", grpc_type="unary", grpc_code!~"DataLoss|DeadlineExceeded|Internal|Unavailable|Unknown"}[10m]))
+          record: central:grpc_server_handled:not_server_error:rate10m
 
         - expr: |
             sum by (namespace, rhacs_instance_id, rhacs_org_id, rhacs_org_name, rhacs_cluster_name, rhacs_environment)
@@ -332,7 +332,7 @@ spec:
         - expr: |
             floor(
               (
-                central:grpc_server_handled:OK:rate10m
+                central:grpc_server_handled:not_server_error:rate10m
               + central:http_incoming_requests:not_5xx:rate10m
               ) / (
                 central:grpc_server_started:total:rate10m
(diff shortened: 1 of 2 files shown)


resources/prometheus/prometheus-rules.yaml
```diff
@@ -307,8 +307,8 @@ spec:
         # GRPC
         - expr: |
             sum by (namespace, rhacs_instance_id, rhacs_org_id, rhacs_org_name, rhacs_cluster_name, rhacs_environment)
-            (rate(grpc_server_handled_total{namespace=~"rhacs-.*", job="central", grpc_type="unary", grpc_code="OK"}[10m]))
-          record: central:grpc_server_handled:OK:rate10m
+            (rate(grpc_server_handled_total{namespace=~"rhacs-.*", job="central", grpc_type="unary", grpc_code!~"DataLoss|DeadlineExceeded|Internal|Unavailable|Unknown"}[10m]))
```
jvdm, line 310: `DataLoss` is "never generated by the library": https://grpc.github.io/grpc/core/md_doc_statuscodes.html

Hence, we would only see it if Central (itself or proxying Scanner) explicitly returns it. Here's my naive check:

```
% find scanner stackrox -type f -name '*.go' | xargs grep -n DataLoss                    
%
```

If we don't explicitly return, is there any reason why introduce it? We could leave it out and assume that all codes listed capture the unavailability scenarios.

Thanks for capturing "Unknown".

resources/prometheus/prometheus-rules.yaml
```diff
@@ -307,8 +307,8 @@ spec:
...
+          record: central:grpc_server_handled:not_server_error:rate10m
```
jvdm, line 311: Nitpick: `not_server_error` -> `server_available_code` or `server_available_status`.

jvdm (commented):
Asking about `DataLoss`.

Also, do you know what gRPC server emits when the server request times out? Just wondering if we should capture that if not captured by the codes we have already in this PR.

resources/prometheus/prometheus-rules.yaml
stehessel: What if we decide to return it in the future? No harm imo to already include it.

jvdm: > What if we decide to return it in the future? 

Given that we don't use it today, we would have to coordinate with Sensor if we decide to use this new status in the future. We might prefer using the other statuses, the ones we are adding (excluding `DataLoss`). I don't have enough info to make that decision right now.

>  No harm imo to already include it.

Generally yes, but I think there are minor drawbacks to adding this status now. First, it's confusing because we don't use it. Why do we have it listed? Second, considering the metric can only be returned if the server explicitly decides to, someone might inadvertently return it, which will impact our availability metric. So unless an immediate reason to add, I think we should just leave it out.

stehessel: Fair enough

jvdm (approved):
LGTM. Thank you for the updates.

PR #97, opened 2023-05-22
ROX-17231 - Alert only on SLO impact

SimonBaeumer:
This pull requests reduces severities of several alerts based on a proposal to only alert during non-working hours for alerts with impact on our SLO.


stehessel (approved):
You need to adapt the other tests as well.

PR #99, opened 2023-05-26
ROX-16561: Add alerts for AWS quotas

vladbologa:


resources/prometheus/prometheus-rules.yaml
```diff
@@ -175,8 +175,69 @@ spec:
...
+          expr: |
+            expr: acs_fleetshard_central_db_instances_used / acs_fleetshard_central_db_instances_max >= 0.9
+          for: 1h
+          labels:
+            severity: critical
+          annotations:
+            summary: "The number of RDS DB instances is very close to its limit."
+            description: "The number of AWS RDS DB instances is close to its limit. A quota increase must be requested from AWS."
+            sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-029-increase-aws-limits.md"
+        - alert: RHACSCentralDBManualSnapshotsUtilizationCritical
```
vladbologa, line 231: Is the inability to delete an instance impact SLOs?

If having an instance stuck in the "deprovisioning" status is OK, this `critical` alert can be removed, and we can only have the `warning` alert.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -176,7 +176,68 @@ spec:
...
+          for: 5m
+          labels:
+            severity: warning
+          annotations:
+            summary: "The number of RDS DB snapshots is close to its limit."
+            description: "The number of AWS RDS DB manual snapshots is close to its limit. A quota increase should be requested from AWS."
+            sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-029-increase-aws-limits.md"
+        - alert: RHACSCentralDBClustersUtilizationCritical
+          expr: |
+            acs_fleetshard_central_db_clusters_used / acs_fleetshard_central_db_clusters_max >= 0.9
```
vladbologa, line 213: Is 0.9 too low for a critical alert? I could change this to 0.95 for example.

Triggering it only when `acs_fleetshard_central_db_clusters_used = acs_fleetshard_central_db_clusters_max` would be too late IMO, as it takes a few hours (up to 4) for AWS to react to a quota increase request.

stehessel: What is the limit? In other words, how many remaining instances do we have left at 90%?

resources/prometheus/prometheus-rules.yaml
```diff
@@ -176,7 +176,68 @@ spec:
...
+          labels:
+            severity: warning
+          annotations:
+            summary: "The number of RDS DB instances is close to its limit."
+            description: "The number of AWS RDS DB instances is close to its limit. A quota increase should be requested from AWS."
+            sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-029-increase-aws-limits.md"
+        - alert: RHACSCentralDBManualSnapshotsUtilizationHigh
+          expr: |
+            acs_fleetshard_central_db_snapshots_used / acs_fleetshard_central_db_snapshots_max >= 0.8
+          for: 5m
```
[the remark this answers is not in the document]
vladbologa: This metric is not influenced by the Probe service creating and deleting instances all the time. The 1h interval is there to prevent the alert firing every time the probe creates an instance (when the current usage is exactly on the threshold).

resources/prometheus/prometheus-rules.yaml
vladbologa: It depends on the current maximum. Which maximum can be increased when the actual usage is close to it, otherwise AWS refuse to increase it. Hence this alert.

Do you think we should rather have this on concrete values than percentages?

E.g. `acs_fleetshard_central_db_clusters_max - acs_fleetshard_central_db_clusters_used < 10`?

stehessel: I think it would be good now how many slots are still left. But we could also add this info (`acs_fleetshard_central_db_clusters_max - acs_fleetshard_central_db_clusters_used`) to the alert description.

vladbologa: Added `acs_fleetshard_central_db_clusters_max - acs_fleetshard_central_db_clusters_used` to the alert description.

resources/prometheus/prometheus-rules.yaml
vladbologa: OK, I'll remove the critical warning.

We request a final snapshot of the DB when deleting it. This is absolutely needed in case of accidental deletion, because the automatic snapshots are all deleted along with the DB. And so, if there's no more quota for manual snapshots, the DB deletion will fail and the ACS instance will be stuck in the deprovisioning state.

stehessel (approved):

PR #103, opened 2023-06-02
Replace HTTP error rate SLI with haproxy metrics

kylape:
These metrics pull from the haproxy instance that fronts all the connections coming in to ACS instances.  This should be more accurate than pulling this SLI from ACS metrics because it will capture requests that never make it to the ACS container for whatever reason (e.g. overloaded, pod is not ready, network issues).

Adjusted instance names and namespaces in unit tests to match the requirement that namespaces look like they do in a real environment (`rhacs-` followed by twenty random alphanumeric characters)


stehessel:
I'm fine with this change in principle. I think there are two things we should do that are missing from this PR:
* Since haproxy does not capture the http traffic from the K8s probes, we should be consistent with grpc and not count `PingService` there as well.
* Without probe traffic, there is now a chance that no requests have been made in the last 10 minutes. As a result, `central:sli:error_rate` would include a zero division/NaN - instead we should count it is `1`.

kylape:
I figured out how to drop all the NaN instances (so they shouldn't alert) by appending `> 0` to the numerator, but I can't seem to force those instances to `1`.  Is there a trick that I'm missing?

stehessel:
No tbh default values and filtering in general are finicky as heck in PromQL. One can use `or vector(1)` to replace NaNs, but then the labels need to match as well. `> 0` is probably good enough, as long as it doesn't cause problems in `central:sli:availability`.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -414,11 +414,11 @@ spec:
                 central:http_incoming_requests:not_5xx:rate10m
               + on (namespace) group_left(rhacs_instance_id, rhacs_org_id, rhacs_org_name, rhacs_cluster_name, rhacs_environment)
                 central:grpc_server_handled:server_available_code:rate10m
-              ) / (
+              ) / ((
                 central:http_incoming_requests:total:rate10m
               + on (namespace) group_left(rhacs_instance_id, rhacs_org_id, rhacs_org_name, rhacs_cluster_name, rhacs_environment)
                 central:grpc_server_started:total:rate10m
-              )
+              ) > 0)
```
stehessel, line 421: Did you check how `central:sli:availability` is affected if there is now a gap in `central:sli:error_rate`? Would there also be a gap in `central:sli:availability`?

kylape: Yes, there would be a gap in those two metrics.  Is that a problem, though?  I tried to create a unit test that showed that the alerts wouldn't fire if this happened, but for some reason the grafana linter tests would fail when trying to parse the dashboard JSON.  Not sure how that was related, but ran out of time.

I went ahead and push the unit test even though I couldn't get the grafana linter to pass in my local box.

stehessel: Yeah, the problem is if the pod is not available (`central:sli:pod_ready=0`), no alerts would fire as well because data points are only present if `central:sli:error_rate` is defined.

stehessel: Perhaps it's possible to do a "left join" where the value of `central:sli:pod_ready` is taken if `central:sli:error_rate` is undefined?

stehessel (approved):

PR #104, opened 2023-06-07
Filter openshift namespaces from CPUThrottlingHigh alert

kylape:
Issues with CPU throttling on openshift namespaces are generally out of our control, and excluding these noisy namespaces will help our signal-to-noise ratio.

### DIFF
--- a/resources/prometheus/kubernetes-mixin-alerts.yaml
+++ b/resources/prometheus/kubernetes-mixin-alerts.yaml
@@ -358,7 +358,7 @@ spec:
             "runbook_url": "https://github.com/kubernetes-monitoring/kubernetes-mixin/tree/master/runbook.md#alert-name-cputhrottlinghigh"
             "summary": "Processes experience elevated CPU throttling."
           "expr": |
-            sum(increase(container_cpu_cfs_throttled_periods_total{container!="", }[5m])) by (container, pod, namespace)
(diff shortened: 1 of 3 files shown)


stehessel (changes requested):
The Kubernetes mixin resources are auto genereated - they should therefore not be modified manually. See also https://github.com/stackrox/rhacs-observability-resources/tree/master/resources/mixins/kubernetes and the `Makefile` therein.

The proper way to modify mixin behavior is to modify the jsonnet configuration in https://github.com/stackrox/rhacs-observability-resources/blob/master/resources/mixins/kubernetes/mixin.libsonnet. In this particular case, you may see from the upstream alert definition that there is a selector variable `cpuThrottlingSelector` that may be set - see https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/003ba5eadfbd69817d1215952133d3ecf99fbd92/alerts/resource_alerts.libsonnet#LL202C87-L202C108. So I would suggest to use this selector to ignore `openshift-*` namespaces.

kylape:
Thanks for the info. I'll take a look next week.

kylape:
Closing this out because the flapping alert this was going to exclude was solved [a different way](https://redhat-internal.slack.com/archives/C055606D477/p1686662872424729?thread_ts=1686633713.044049&cid=C055606D477).

PR #108, opened 2023-06-22
Generate kube-state alarms only for RHACS namespace

jvdm:
Set `namespaceSelector: 'namespace=~"rhacs-.*|acscs-.*"'` in jsonnet `_config`, to effectively have alerts only on ACS workloads.

See [the related thread](https://redhat-internal.slack.com/archives/CCX9DB894/p1687381918802879?thread_ts=1687297268.055499&cid=CCX9DB894) for the context on why SRE supports this.

### DIFF
--- a/resources/mixins/kubernetes/mixin.libsonnet
+++ b/resources/mixins/kubernetes/mixin.libsonnet
@@ -9,7 +9,7 @@ kubernetes {
     kubeApiserverSelector: 'job="api"',
     kubeProxySelector: 'job="machine-config-daemon"',
     kubeSchedulerSelector: 'job="scheduler"',
-    namespaceSelector: 'namespace=~"rhacs-.*|acscs-.*"',
+    namespaceSelector: 'namespace=~"rhacs|rhacs-.*|acscs-.*"',
   },
 } + {
   // Customize alert labels.
--- a/resources/mixins/kubernetes/mixin.libsonnet
+++ b/resources/mixins/kubernetes/mixin.libsonnet
@@ -9,7 +9,7 @@ kubernetes {
     kubeApiserverSelector: 'job="api"',
     kubeProxySelector: 'job="machine-config-daemon"',
     kubeSchedulerSelector: 'job="scheduler"',
-    kubeStateMetricsSelector: 'namespace!~"openshift-.*',
+    namespaceSelector: 'namespace=~"rhacs-.*|acscs-.*"',
   },
 } + {
   // Customize alert labels.
--- a/resources/mixins/kubernetes/mixin.libsonnet
+++ b/resources/mixins/kubernetes/mixin.libsonnet
@@ -9,6 +9,7 @@ kubernetes {
     kubeApiserverSelector: 'job="api"',
     kubeProxySelector: 'job="machine-config-daemon"',
     kubeSchedulerSelector: 'job="scheduler"',
+    kubeStateMetricsSelector: 'namespace!~"openshift-.*',
   },
 } + {
   // Customize alert labels.
--- a/resources/mixins/kubernetes/mixin.libsonnet
+++ b/resources/mixins/kubernetes/mixin.libsonnet
@@ -9,7 +9,7 @@ kubernetes {
     kubeApiserverSelector: 'job="api"',
     kubeProxySelector: 'job="machine-config-daemon"',
     kubeSchedulerSelector: 'job="scheduler"',
-    namespaceSelector: 'namespace=~"rhacs|rhacs-.*|acscs-.*"',
+    namespaceSelector: 'namespace!~"openshift-kube.*|kube.*"'
   },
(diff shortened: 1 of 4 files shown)


resources/mixins/kubernetes/mixin.libsonnet
```diff
@@ -9,6 +9,7 @@ kubernetes {
     kubeApiserverSelector: 'job="api"',
     kubeProxySelector: 'job="machine-config-daemon"',
     kubeSchedulerSelector: 'job="scheduler"',
+    namespaceSelector: 'namespace=~"rhacs-.*|acscs-.*"',
```
porridge, line 12: I also found the following possibly interesting and one definitely interesting namespace in `acs-stage-dp-02` that don't match your regex:
- `dinosaur-counter`
- `external-secrets`
-  `rhacs`

porridge (changes requested):

resources/mixins/kubernetes/mixin.libsonnet
jvdm: How can we confirm which namespaces we want to monitor (except the obvious `rhacs`)?

```
% kubectl get ns | grep -Ev '^(openshift|kube|rhacs-|acscs-)' 
NAME                                               STATUS   AGE
dedicated-admin                                    Active   216d
default                                            Active   216d
dinosaur-counter                                   Active   195d
external-secrets                                   Active   19d
qa                                                 Active   125d
rhacs                                              Active   202d
```

jvdm: Should we blocklist instead, e.g.: `namespace!~"openshift-.*|kube-.*"`?

porridge: I'm 100% sure we want `rhacs`. @kovayur would know about `external-secrets`.
I don't have any good ideas about the rest other than asking on slack.

However this highlights an important point - how do we make sure going forward that this list is up to date as the system evolves? Maybe a denylist is indeed a more robust solution. Better to have to add to it than miss something important :thinking:

kurlov: I think it makes sense to ignore `openshift-.*` and `kube-.*` namespaces. It's highly unlikely to use those for new deployment so no need to think about keeping the list up to date

stehessel: I think even `openshift-.*` might be too broad. For example, we use OpenShift logging operator for ACS-CS logging, which is installed in `openshift-logging`. Another candidate would be `openshift-authentication`, which we use for authn/z to access the cluster. I would suggest to start with a conservative deny list that takes care of the alerts that annoyed us so far, and expand it as we go.

jvdm: @porridge:

> how do we make sure going forward that this list is up to date as the system evolves?

If we have control over the name of the namespaces we will need, I would say an allowlist has the benefit of the namespace creator to decide if he wants to have his workload monitored or not (by prefixing `rhacs-` or `acscs-`).

jvdm: @stehessel:

> I think even openshift-.* might be too broad. For example, we use OpenShift logging operator for ACS-CS logging, which is installed in openshift-logging

The idea behind blocking `openshift-` is that SRE already has monitoring that covers platform workloads.   That being said, SRE does not monitor every pod.  They have symptom-based alerts that *should* cover issues with platform components that will direct to the specific causes.

> I would suggest to start with a conservative deny list that takes care of the alerts that annoyed us so far, and expand it as we go.

Maybe we should think if we currently have any actionable steps for the `openshift-` alarms when they are triggered and not impacting ACSCS. Do you think it would be more practical to use alarms that monitor the impact in ACSCS instead?

I agree that some of the `openshift-` alarms might be interesting to monitor, but these would be the exceptions, AFAIK. We have ~100 openshfit namespaces. Also, I think the static pods that annoyed us might exist in multiple namespaces. In the latest alarm they were all `openshift-kube-.*`, though. So maybe block that instead for now, with `kube-.*`.

jvdm: Moving to blocklist as suggested by me above, we have the following list in `eu-01`:

```
% kubectl get ns | grep -E '^(openshift-kube|kube)' | awk '{print $1}'
kube-node-lease
kube-public
kube-system
openshift-kube-apiserver
openshift-kube-apiserver-operator
openshift-kube-controller-manager
openshift-kube-controller-manager-operator
openshift-kube-scheduler
openshift-kube-scheduler-operator
openshift-kube-storage-version-migrator
openshift-kube-storage-version-migrator-operator
%
```

stehessel (approved):
LGTM. By the way, @porridge seems to be on PTO. But I think his concerns were addressed, so I would say feel free to override his request for change.

PR #109, opened 2023-06-26
Increase paused central alert warning duration

SimonBaeumer:
Increase the duration of the paused instance alert to 30 days.
We expect a release every 9 weeks, therefore alerting after two days is too early to alert a warning when no fix was released for ACS.

### DIFF
--- a/resources/prometheus/unit_tests/RHACSFleetshardSyncCentralReconcilePaused.yaml
+++ b/resources/prometheus/unit_tests/RHACSFleetshardSyncCentralReconcilePaused.yaml
@@ -12,10 +12,10 @@ tests:
       - eval_time: 1h
         alertname: RHACSFleetshardSyncCentralReconcilePaused
         exp_alerts: []
-      - eval_time: 30h
+      - eval_time: 20d
         alertname: RHACSFleetshardSyncCentralReconcilePaused
         exp_alerts: []
-      - eval_time: 49h
+      - eval_time: 31d
         alertname: RHACSFleetshardSyncCentralReconcilePaused
         exp_alerts:
           - exp_labels:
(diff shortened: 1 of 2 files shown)


resources/prometheus/unit_tests/RHACSFleetshardSyncCentralReconcilePaused.yaml
```diff
@@ -12,18 +12,18 @@ tests:
...
         exp_alerts:
           - exp_labels:
               alertname: RHACSFleetshardSyncCentralReconcilePaused
               instance: rhacs-chs64i3dabr0026a6fag
               severity: warning
             exp_annotations:
-              summary: "ACS instance rhacs-chs64i3dabr0026a6fag has paused reconciliation for more than 2 days."
+              summary: "ACS instance rhacs-chs64i3dabr0026a6fag has paused reconciliation for more than 30 days."
               description: "ACS instance rhacs-chs64i3dabr0026a6fag has the 'pause-reconcile' annotation and therefore is not being managed by the ACS operator. Please check that this is intended."
       - eval_time: 52h
```
SimonBaeumer, line 28: Should this be increased to 32 days so it tests a silenced alert?

resources/prometheus/unit_tests/RHACSFleetshardSyncCentralReconcilePaused.yaml
```diff
@@ -12,18 +12,18 @@ tests:
       - eval_time: 1h
         alertname: RHACSFleetshardSyncCentralReconcilePaused
         exp_alerts: []
-      - eval_time: 30h
```
stehessel, line 15: You need to add more values to the test series, otherwise there is no data at `eval_time`.

vladbologa (changes requested):
This alert will never fire if we set `for: 30d`

Because the metric is provided by `fleetshard-sync`, it's not provided during ACSCS releases (usually every 14 days) and basically the interval will reset to 0 every 14 days (or every time `fleetshard-sync` restarts)

stehessel:
Hmm the `sum by (exported_instance)` will resolve the issue of fleetshard-sync's pod label changing, but I think the release can still reset the `for` timer if there is a period without metrics.

We could try to solve this by using something like `avg_over_time` for 30 days, but it would have quite bad query performance. I would actually propose to remove the alert all together and instead add this to query to a dashboard.

SimonBaeumer:
@stehessel Thank for helping me out! Indeed,  I agree it is better to have it as a dashboard in Grafana. I am closing this PR in favor of the dashboard.

stehessel:
Alright, let's remove the alert then as well. FYI there is another open PR that I'd like to merge before promoting new observability changes to prod - https://github.com/stackrox/rhacs-observability-resources/pull/112.

PR #111, opened 2023-06-29
Revert "Merge pull request #110 from stackrox/master"

stehessel:
This reverts commit 807fb04ddf614b263eb31c1153df5eaa663fd37a, reversing changes made to 3f2b1531f2192dd7561227b01a351777412183be.

Need to revert the haproxy SLI changes, because they break the prometheus labels (no `rhacs_*` labels). As a result, the dashboards are broken as well.

PR #112, opened 2023-06-29
fix labels in SLI metrics

stehessel:
The issue addressed here is that with the latest haproxy metrics changes, we are running into scenarios where metrics are undefined. In these cases we provide defaults, which however do not have the correct `rhacs_*` labels.

To fix this issue, we always add `rhacs_instance_id` to `central:sli:pod_ready`. This helps us to make sure that `central:sli:error_rate` has either all `rhacs_*` labels or at least `rhacs_instance_id` attached. This should prevent Prometheus from dropping metric series or messing up `central:sli:availability`.

To summarize, the following scenarios may occur:
* `>= 1` pods available, but no incoming requests
  * all error rate metrics are undefined.
  * `central:sli:error_rate` defaults to `rox_central_postgres_connected=1`.
  * `rhacs_instance_id`, `rhacs_org_id`, `rhacs_org_name`, `rhacs_cluster_name`, `rhacs_environment` labels are attached.
  * This is the more common scenario where undefined metrics may arise.
* `0` pods available
  * all error rate metrics are undefined.
  * `central:sli:error_rate` defaults to `central:sli:pod_ready=0`.
  * `namespace` and `rhacs_instance_id` labels are attached.
* `>= 1` pods available with incoming requests
  * happy path, all `rhacs_*` labels are attached. No undefined metric series.

This should fix the issue observed in https://github.com/stackrox/rhacs-observability-resources/pull/111.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -336,7 +336,15 @@ spec:
         # Central pod must be ready to count as available.
         # This is a time series of 0s (down) and 1s (up).
         - expr: |
-            clamp_max(sum by(namespace) (kube_deployment_status_replicas_ready{deployment="central", namespace=~"rhacs-.*"}), 1)
+            label_replace(
+              clamp_max(
+                sum by(namespace) (kube_deployment_status_replicas_ready{deployment="central", namespace=~"rhacs-.*"}), 1
+              ),
+              "rhacs_instance_id",
+              "$1",
+              "namespace",
+              "rhacs-(.*)"
+            )
           record: central:sli:pod_ready
 
         # Error rate SLI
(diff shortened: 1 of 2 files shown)


kylape (commented):
Overall the changes look reasonable, though I haven't tested them yet.  Perhaps we can merge this and test the scenarios out in stage?

Couple other questions:

1. Do you think we should add more unit tests for the scenarios you listed in your PR description?
2. I know you said that rhacs_instance_id is sufficient for dashboards, but I can't think of the other scenarios where having the extra labels from the fallback metric rox_central_postgres_connected are valuable.  Is it for alerts?

stehessel:
> Overall the changes look reasonable, though I haven't tested them yet. Perhaps we can merge this and test the scenarios out in stage?

Yeah I think that would be a good idea. Unfortunately the interaction between metrics and dashboards can be subtle.

> Couple other questions:
> 
> 1. Do you think we should add more unit tests for the scenarios you listed in your PR description?

Not a bad idea, I can look into adding some more testing.

> 2. I know you said that rhacs_instance_id is sufficient for dashboards, but I can't think of the other scenarios where having the extra labels from the fallback metric rox_central_postgres_connected are valuable.  Is it for alerts?

Well, `rhacs_instance_id` is *mostly* sufficient - there will be data in widgets which filter for the instance like with the query `avg(central:sli:availability{rhacs_instance_id=~"$instance_id"})`. However, we also have a table view where the organisation is displayed (which is very useful). The query is `central:sli:availability:extended_avg_over_time28d{rhacs_org_id!="16536854"}`, which would not show the organisation in cases when we default to `sli:pod_ready`. Maybe we could work around that with some more joins in the dashboard query, but in general I'd prefer to have all the labels whenever we can. Sadly I did not find a way to get the labels when Central itself is down, because the labels come from Central.

Another example would simply be an adhoc promql query for the SLI by organisation. With these changes, at least if the pod is available this would always work.

stehessel:
So I slept on this and kind of changed my mind. While I do think all the labels are useful for the reasons above, I'm concerned by the complexity in the promQL to make it happen. Also metrics with changing labels are treated as new metrics series by Prometheus, so one has to be very careful with the queries.

So I'd say let's maybe start with just `rhacs_instance_id` on all of the SLI metrics to be consistent, add see how we can deal with that on the Grafana side.

stehessel:
I'll build back the changes to align with ^.

stehessel:
@kylape Could you take another look? As mentioned I decided to move some of the more complicated labelling logic into the dashboards. I'm not 100% that will work, but we need at consistent `rhacs_instance_id` labels to try, which is what this PR now provides.

kylape (approved):
changes look good to me.  I agree simpler promql is better :relieved:

PR #113, opened 2023-07-04
ROX-18265: remove pause reconcile alert

stehessel:
* Remove the paused reconcile alert. It seems to do more harm than good and is often not actionable for on-call. Also increasing the pending duration >2 weeks is tricky because then the alert is not triggered at all due to releases.
* Add a widget to the cluster overview dashboard that shows the paused instance count, and add a paused column to the instance table.

Screenshot 2023-07-04 at 14 53 46
Screenshot 2023-07-04 at 14 54 02


kurlov (approved):
I was about to implement the same 😄

PR #115, opened 2023-07-17
use max values over time in probe alert

stehessel:
This addresses an issue with the current probe alerting condition. If the probe fails from the very beginning when the probe pod is live, `acs_probe_last_success_timestamp` is never defined and the condition does not evaluate to true. To fix this, we look back a week and find the last recorded timestamps of any probe pod.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -236,13 +236,16 @@ spec:
     - name: rhacs-probe
       rules:
         - alert: RHACSProbeRunFailed
-          expr: acs_probe_last_failure_timestamp > 0 and acs_probe_last_failure_timestamp >= acs_probe_last_success_timestamp
+          expr: |
+            max by (region, rhacs_cluster_name, rhacs_environment) (max_over_time(acs_probe_last_failure_timestamp[1w]))
+              >
+            max by (region, rhacs_cluster_name, rhacs_environment) (max_over_time(acs_probe_last_success_timestamp[1w]))
           for: 30m
           labels:
             severity: critical
           annotations:
             summary: "The latest probe run failed at `{{ $value | humanizeTimestamp }}`."
-            description: "The latest run of probe `{{ $labels.pod }}` failed at `{{ $value | humanizeTimestamp }}`."
+            description: "The latest probe run in region `{{ $labels.region }}` on cluster `{{ $labels.rhacs_cluster_name }}` failed at `{{ $value | humanizeTimestamp }}`."
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-008-probe-run-failed.md"
         - alert: RHACSProbeScrapeFailed
           expr: |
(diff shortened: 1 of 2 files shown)


resources/prometheus/unit_tests/RHACSProbeRunFailed.yaml
```diff
@@ -6,10 +6,10 @@ evaluation_interval: 1m
...
   - interval: 1m
     input_series:
-      - series: acs_probe_last_failure_timestamp{namespace="rhacs-probe", pod="probe-1234"}
-        values: "0+0x10 0+0x15 7+0x60"
-      - series: acs_probe_last_success_timestamp{namespace="rhacs-probe", pod="probe-1234"}
-        values: "0+0x10 1+1x15 6+0x60"
+      - series: acs_probe_last_failure_timestamp{namespace="rhacs-probe", region="eu-west-1", rhacs_cluster_name="acs-eu-1"}
+        values: "0+0x15 17+1x60"
+      - series: acs_probe_last_success_timestamp{namespace="rhacs-probe", region="eu-west-1", rhacs_cluster_name="acs-eu-1"}
+        values: "1+1x15 17+0x60"
```
janisz, line 12: why this has changed?

resources/prometheus/prometheus-rules.yaml
```diff
@@ -236,13 +236,16 @@ spec:
     - name: rhacs-probe
       rules:
         - alert: RHACSProbeRunFailed
-          expr: acs_probe_last_failure_timestamp > 0 and acs_probe_last_failure_timestamp >= acs_probe_last_success_timestamp
+          expr: |
+            max by (region, rhacs_cluster_name, rhacs_environment) (max_over_time(acs_probe_last_failure_timestamp[1w]))
+              >
+            max by (region, rhacs_cluster_name, rhacs_environment) (max_over_time(acs_probe_last_success_timestamp[1w]))
```
janisz, line 242: 1w sounds like quite long time. Why not 24h?

I think I need more explanation. I don't understand this query. We take max timestamp from last week and compare them.

resources/prometheus/unit_tests/RHACSProbeRunFailed.yaml
stehessel: Because the old values were actually an incorrect counter series (not monotonous). Didn't matter before, but does matter now.

resources/prometheus/prometheus-rules.yaml
stehessel: Note that we are dealing with timestamps here, which are monotonous. So taking the max represents the latest timestamp.

> I think I need more explanation. I don't understand this query. We take max timestamp from last week and compare them.

In planin English we are trying to do this: If either the last success or the last failure timestamp are not defined, use the most recent timestamp during the last week.

Why one week? It's arbitrary, but I thought it a reasonable compromise between going back to the end of time (which we logically would like), query performance and the probability that there is at least one undefined value in the period.

stehessel:
I decided to also implement the initial zero values for the timestamp gauges directly in the probe, see https://github.com/stackrox/acs-fleet-manager/pull/1162. I believe that will also fix the issue with the need for this PR, although it's not 100% equivalent. The main difference being that without this PR, re-deploying the probe would still temporarily reset the alert.

janisz (approved):

stehessel:
I'll keep this PR open for bit to see if it is still necessary, or if the problem was sufficiently fixed by https://github.com/stackrox/acs-fleet-manager/pull/1162.

PR #117, opened 2023-07-20
ROX-17469: implemented sli/alerts for central api latencies 

pepedocs:
**What**: Draft PR for implementation of latency SLI and alerts for central API.

**The acceptance criteria is as follows:**
- Latency included in SLO dashboard
- Critical SLI alerts, following the pattern of other SLI alerts
- Documented

**Additional Information:** It was suggested that different API types' latencies are grouped for simplicity reasons.

**How (Proposal)**
1. Separate SLIs for each API type.
  - Can be instantly used from an incident management/troubleshooting perspective.
  - The underlying queries are much simpler and thus easier to read.

2. Only implement the SLI `"Only 90% of the requests served under 100ms"` for now.
  - Lesser SLIs
  - The other SLI `"Only 99% of the requests served under 1000ms"` is more likely to produce false positives.
  - I do not know if there is an immediate need for this SLI.

**Tests**
   - Added unit-tests
   - Testing alerts/queries with Prometheus's histogram_quantile function is not straightforward and can be outdated when the histogram buckets are updated. Therefore, instead of testing the alerts, the alerts' intermediate components are tested as they are much simpler to test.


stehessel (commented):
> Separate SLIs for each API type.

I think this is fine, but we should be careful with increasing the SLI cardinality, otherwise we will end up with a huge matrix of SLI metrics. I agree with skipping the p99 SLI for this reason, although I remember that others felt strongly about these (+ even more like p100 ...).

#### Alerting

On a more general note, I would like to see the alerting based on error budgets and burn rates - see `central:slo:availability:error_budget_exhaustion` and `central:slo:availability:burnrate1h` as examples. I see two strategies on how to achieve that for latency:

1. **Time based SLI**

The SLO period (28 days) is divided into time chunks, in which the SLI is either met or violated. This is how we calculate the availability SLI. In the case of availability, it means if the service is determined to be up or down. For latency, we could use 10 minute time chunks similar to what you have defined, and count how much time the SLI is met. This allows us to transcribe the latency into the "9 notation":

`99% of the time, the 10 minute latency p90 must be < 100ms. We are currently at 99.5% over the last month. The error budget is at 50%.`

As a side note, we could even roll the latency SLI into the availability SLI. The interpretation would be: "If the service latency was too high for the last 10 minutes, it was unavailable." However, I would argue against this because that is not how availability is defined in our SLAs, and it would complicate our metrics definition.

2. **Counter based SLI**

We measure p90 over the entire SLO period of 28 days. The transcription would be:

`p90 must be < 100ms. We are currently at p90=50ms over the last 28 days. The error budget is at 50%.`

#### Other random thoughts

* The Central traffic is actually quite low in many cases (<1 rps). In some cases it may also be zero (e.g. no sensor is connected, nobody is using the UI, ...). This makes it hard to get good statistics for the percentiles.
* Looking at the graphql latencies, I don't think 100ms is a reasonable target unfortunately. I believe that is because some of the UI queries actually scale with the data (e.g. number of deployments in the cluster), so they may take quite some time. As far as I know the UI is rendered synchronously. Obviously hanging UIs are bad user experience, but in regards to SLIs, that may be expected for some queries. Perhaps you could talk to the UI team (https://redhat-internal.slack.com/archives/C7ERNFL0M) about what the general latency expectation is.
* The grpc metrics as defined include internal endpoints (e.g. `PingService`), which are lumped in actual Sensor/roxctl requests. We may choose to ignore this for simplicity, but it's something to be aware of when interpreting the data. Concretely, I expect this to skew the latencies to lower end.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -454,6 +454,78 @@ spec:
...
+        - expr: |
+            1 - (sum_over_time(central:grpc_server_handling_seconds:rate10m:p90[1h]) / count_over_time(central:grpc_server_handling_seconds:rate10m:p90[1h]))
+          record: central:grpc_server_handling_seconds:rate10m:p90:error_rate1h
+        - expr: |
+            1 - (sum_over_time(central:http_incoming_request_duration_histogram_seconds_bucket:rate10m:p90[1h]) / count_over_time(central:http_incoming_request_duration_histogram_seconds_bucket:rate10m:p90[1h]))
+          record: central:http_incoming_request_duration_histogram_seconds_bucket:rate10m:p90:error_rate1h
+
+        # - Queries the error budget exhaustion (or consumption) for the whole slo window (28d).
+        - expr: |
+            (1 - central:grpc_server_handling_seconds:rate10m:p90:sli) / 0.01
```
stehessel, line 498: Can you define a scalar recording rule for the target (`0.01` / `0.99`)? That allows us to change the value in a single place.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -639,3 +716,96 @@ spec:
...
+          labels:
+            service: central
+            namespace: "{{ $labels.namespace }}"
+            rhacs_instance_id: "{{ $labels.namespace }}"
+            path: "{{ $labels.path }}"
+            severity: warning
+        - alert: Central latency burn rate for GRPC API
+          annotations:
+            message: "Latency  burn rate for central's GRPC API. Current burn rate per hour: {{ $value | humanize }}."
+          expr: |
```
stehessel, line 792: Why did you chose `0.5` as the burn rate threshold? That seems very low to me. Note that by definition, a burn rate of 1 means that the full error budget will be consumed after 28 days.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -454,6 +454,78 @@ spec:
             )
           record: central:sli:availability:extended_avg_over_time28d
 
+        # - Queries the 90th percentile of central's handled GRPC/HTTP API requests latencies over the last 10 minutes.
+        - expr: |
+            (histogram_quantile(0.9, sum  by(le, namespace, grpc_service, grpc_method) (rate(grpc_server_handling_seconds_bucket{container="central", grpc_method!~"ScanImageInternal|DeleteImages|EnrichLocalImageInternal|RunReport|ScanImage|TriggerExternalBackup|Ping"}[10m]))) > 0) < bool 0.1
```
stehessel, line 459: TIL, didn't know about `< bool`. Probably can use that else where as well to simplify the promQL.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -454,6 +454,78 @@ spec:
...
+            sum_over_time(central:grpc_server_handling_seconds:rate10m:p90[28d]) / count_over_time(central:grpc_server_handling_seconds:rate10m:p90[28d])
+          record: central:grpc_server_handling_seconds:rate10m:p90:sli
+        - expr: |
+            sum_over_time(central:http_incoming_request_duration_histogram_seconds_bucket:rate10m:p90[28d]) / count_over_time(central:http_incoming_request_duration_histogram_seconds_bucket:rate10m:p90[28d])
+          record: central:http_incoming_request_duration_histogram_seconds_bucket:rate10m:p90:sli
+
+        # - Queries the error rate or the ratio of the instances of central:xxx:rate10m:p90
+        #   were equal to 0 over the total instances of central:xxx:rate10m:p90 within a period.
+        - expr: |
+            1 - central:grpc_server_handling_seconds:rate10m:p90:sli
```
stehessel, line 482: Is there a reason you changed the order in the naming compared to existing metrics? E.g. `central:grpc_server_handling_seconds:rate10m:p90:sli` vs `central:sli:grpc_server_handling_seconds:rate10m:p90`.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -454,6 +454,78 @@ spec:
...
+        # - Queries error budget burn rate (a.k.a. burn rate) is the ratio of central:xxx:rate10m:p90:error_rateyyy
+        #   over the error budget for a period (e.g. 1h, 1d, etc).
+        - expr: |
+            central:grpc_server_handling_seconds:rate10m:p90:error_rate1h / 0.01
+          record: central:grpc_server_handling_seconds:rate10m:p90:burn_rate1h
+        - expr: |
+            central:http_incoming_request_duration_histogram_seconds_bucket:rate10m:p90:error_rate1h / 0.01
+          record: central:http_incoming_request_duration_histogram_seconds_bucket:rate10m:p90:burn_rate1h
+
+        # - A sample count filter that ignores central:xxx:rate10m:p90 instances that has samples less than the expected sample count.
```
stehessel, line 513: How does the filter treat periods where there is no incoming traffic (and the base metrics are therefore undefined)?

resources/prometheus/prometheus-rules.yaml
```diff
@@ -533,6 +605,11 @@ spec:
             severity: critical
             namespace: "{{ $labels.namespace }}"
             rhacs_instance_id: "{{ $labels.rhacs_instance_id }}"
+            rhacs_org_name: "{{ $labels.rhacs_org_name }}"
```
stehessel, line 608: We can't add these labels here in general, because the values originate in Central, and if Central itself is down, they don't exist.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -639,3 +716,96 @@ spec:
...
+
+        - alert: Central latency error budget exhaustion for GRPC API - 90%
+          annotations:
+            message: "Latency error budget exhaustion for central's GRPC API. Current exhaustion: {{ $value | humanizePercentage }}."
+          expr: |
+            (central:grpc_server_handling_seconds:rate10m:p90:sample_count_filter28d * central:grpc_server_handling_seconds:rate10m:p90:error_budget_exhaustion28d) >= 0.9
+          labels:
+            service: central
+            namespace: "{{ $labels.namespace }}"
+            rhacs_instance_id: "{{ $labels.namespace }}"
```
stehessel, line 728: The instance id is not the same as the namespace. The namespace is `rhacs-{instance_id}`.

resources/prometheus/prometheus-rules.yaml
stehessel, line 459: The incoming gRPC calls are already very sparse for most Centrals. I think we should consider consolidating them if they roughly do the same thing latency wise. So I would sum here over `grpc_service` / `grpc_method`.

stehessel:
In general looks good to me, but I think we should plan some time to observe the metrics in stage and maybe iterate the values if needed.

PR #118, opened 2023-07-20
fix labels in SLO dashboard

stehessel:
Follow up to #112. Since there are now only `namespace` and `rhacs_instance_id` labels on the SLI metrics like `sli:central:availability`, we need to use a different metric for the dashboard variables. We also need to merge another metric (`process_cpu_seconds_total`) onto the table data to get the `org_name` and `org_id` columns.


resources/grafana/rhacs-central-slo-dashboard.yaml
```diff
@@ -162,7 +162,7 @@ spec:
               },
               "editorMode": "code",
               "exemplar": false,
-              "expr": "count(central:slo:availability:error_budget_exhaustion{rhacs_org_id!=\"16536854\"} <= 1 and central:sli:availability) or vector(0)",
+              "expr": "count(central:slo:availability:error_budget_exhaustion{rhacs_instance_id=~\"$instance_id\"} <= 1 and central:sli:availability) or vector(0)",
```
ivan-degtiarenko, line 165: What's `$instance_id` and where is the value coming from? And why this change(and similar changes) here is needed?

resources/grafana/rhacs-central-slo-dashboard.yaml
```diff
@@ -521,35 +536,73 @@ spec:
...
-                  "Time 2": 11,
+                  "Time 1": 9,
+                  "Time 2": 10,
+                  "Time 3": 13,
                   "Value #A": 1,
                   "Value #B": 2,
-                  "__name__": 5,
-                  "namespace": 7,
-                  "observability": 8,
+                  "Value #C": 28,
```
ivan-degtiarenko, line 571: Are these generated changes? If not, what is the purpose behind them? Talking about lines https://github.com/stackrox/rhacs-observability-resources/pull/118/files#diff-cec7c1223ca9d5671a37727817b570e4bc72df0cd59f33280bb7e5656d354020R539-R605

ivan-degtiarenko (commented):
Some questions for me to understand what's happening better

resources/grafana/rhacs-central-slo-dashboard.yaml
stehessel: `$instance_id` is part of the dashboard variables you see on top of the screen. Note that multiple instances may be selected - note that `=~`, which in promql means compare with regex.

Screenshot 2023-07-25 at 00 26 04

These variables work like this: You can select a metric, and it gives you all label values of this metric for selection. Previously we used something like `central:sli:availability{org_id!="probe-organisation"}` as the metric.

Screenshot 2023-07-25 at 00 29 13

Now the major change that was fixed in https://github.com/stackrox/rhacs-observability-resources/pull/112 is that SLI metrics no longer have `rhacs_org_*` labels. That is because some of the SLI now come from HAProxy, which has no knowledge about ACS tenants. As a result `central:sli:availability` and derived metrics also no longer have `rhacs_org_*` labels. We therefore need to change the variable definition to another metric (`process_cpu_seconds_total`). We also need to merge this metric in the table view to get all the columns.

resources/grafana/rhacs-central-slo-dashboard.yaml
stehessel: The entire json is generated, however it's based on the changes that I made via the Grafana UI. The process is basically like this:
* Load the dashboard in Grafana.
* Make the changes you want, like updating/adding/removing widgets.
* Export the entire dashboard as json and paste it into the PR.

Specifically, what you are looking at here is the diff generated by adding another query to the table view. You can select all labels of the result as columns in the table. In this case I hid most of them and only select org name and org id.

ivan-degtiarenko (approved):
Thanks for the clarification - LGTM 👍

PR #119, opened 2023-07-21
Ignore kube-mixin alerts for openshift-marketplace

dhaus67:
# Description

Increasing the area of ignored alerts due to some unwanted pages for the `openshift-marketplace` namespace (in this case the `community-operator` pod).

_Note for reviewers: unsure why there are so many changes within the dashboard after running `make generate`._

### DIFF
--- a/resources/mixins/kubernetes/mixin.libsonnet
+++ b/resources/mixins/kubernetes/mixin.libsonnet
@@ -9,7 +9,7 @@ kubernetes {
     kubeApiserverSelector: 'job="api"',
     kubeProxySelector: 'job="machine-config-daemon"',
     kubeSchedulerSelector: 'job="scheduler"',
-    namespaceSelector: 'namespace!~"openshift-kube.*|kube.*"'
+    namespaceSelector: 'namespace!~"openshift-kube.*|openshift-marketplace|kube.*"'
   },
 } + {
   // Customize alert labels.
--- a/resources/grafana/mixins/kubernetes/k8s-resources-cluster.yaml
+++ b/resources/grafana/mixins/kubernetes/k8s-resources-cluster.yaml
@@ -549,8 +549,7 @@ spec:
                   "format": "time_series",
                   "intervalFactor": 2,
                   "legendFormat": "{{namespace}}",
-                  "legendLink": null,
-                  "step": 10
+                  "legendLink": null
                 }
               ],
               "thresholds": [],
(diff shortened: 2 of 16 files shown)


resources/mixins/kubernetes/mixin.libsonnet
```diff
@@ -9,7 +9,7 @@ kubernetes {
     kubeApiserverSelector: 'job="api"',
     kubeProxySelector: 'job="machine-config-daemon"',
     kubeSchedulerSelector: 'job="scheduler"',
-    namespaceSelector: 'namespace!~"openshift-kube.*|kube.*"'
+    namespaceSelector: 'namespace!~"openshift-kube.*|openshift-marketplace|kube.*"'
```
stehessel, line 12: Looking over the `openshift-*` namespaces, should we just ignore everything except maybe `openshift-monitoring`, `openshift-authentication`, `openshift-ingress` and `openshift-logging`? Those are probably the most important ones for us that we would care about even if OSD manages them.

stehessel:
> Note for reviewers: unsure why there are so many changes within the dashboard after running make generate.

The mixin version got updated because of

```
.PHONY: generate
generate: update alerts dashboards
```

I guess we could separate `generate` and `update` to prevent this from happening.

resources/mixins/kubernetes/mixin.libsonnet
dhaus67: Hmm, I guess we could do that. I'm wondering whether we are now at a point where we can switch to a allowlist instead of a deny list after [the initial discussion here.](https://github.com/stackrox/rhacs-observability-resources/pull/108#discussion_r1237980359)

Looking at prod, the most relevant namespaces seem to be:
```
acscs-dataplane-cd
openshift-authentication
openshift-authentication-operator
openshift-ingress
openshift-ingress-canary
openshift-ingress-operator
openshift-logging
openshift-monitoring
rhacs
rhacs-<id>
rhacs-cloudwatch
rhacs-observability
rhacs-probe
```

which would be included by `namespaceSelector: 'namespace~"acscs.*|openshift-authentication.*|openshift-ingress.*|openshift-logging|openshift-monitoring|rhacs.*"'`.

stehessel (approved):
The proposed deny list looks reasonable to me. But we can also let it soak a bit longer and revisit this topic later, it doesn't seem like the alert volume is a big problem currently.

PR #120, opened 2023-07-27
added rhacs-consolidated.json and scripts to generate yamls

vinaybommana:
Hey folks added consolidated dashboard json which consists of RHACS cluster overview and SLO dashboards, the idea behind this is to have a single source of truth for in-cluster [dashboards](https://grafana-route-rhacs-observability.apps.acs-prod-dp-01.pnz3.p1.openshiftapps.com/?orgId=1) and consolidated [dashboards](https://grafana.app-sre.devshift.net/d/I3Ms3kuVz/rhacs-consolidated-dashboard?orgId=1), I've already added the consolidated dashboard in app-interface having the yaml at managed-tenants-slos [repo](https://gitlab.cee.redhat.com/service/managed-tenants-slos/-/blob/main/RHACS/dashboards/rhacs-consolidated.configmap.yaml) this PR helps to maintain the consolidated dashboards at single source. 
After the PR merge I'll change the saas [file](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/data/services/addons/cicd/dashboards/saas-ms-rhacs-dashboards.yaml) in app-interface to include this repo

created the following directory structure same as mixins 
Screenshot from 2023-07-27 11-56-59
the dashboard yamls are generated the same way the mixin dashboards will `make generate`
please suggest changes if I am missing something,  the dashboard I've included can be viewed at [RHACS consolidated dashboard](https://grafana.app-sre.devshift.net/d/I3Ms3kuVz/rhacs-consolidated-dashboard?orgId=1)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
 .PHONY: generate
 generate:
 	$(MAKE) -C resources/mixins/kubernetes generate
+	$(MAKE) -C resources/grafana generate
--- a/resources/grafana/Makefile
+++ b/resources/grafana/Makefile
@@ -1,0 +1,6 @@
+.PHONY:
+dashboards: .
+	@scripts/generate-dashboards.sh
+
+.PHONY: generate
+generate: dashboards
--- a/resources/grafana/generated/dashboards/rhacs-consolidated.json
+++ b/resources/grafana/generated/dashboards/rhacs-consolidated.json
@@ -1,0 +1,3015 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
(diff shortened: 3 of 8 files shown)


stehessel:
This goes into the right direction, however there are still some problems:

* The idea would to replace the existing dashboards (cluster overview, cluster SLO) with the generated json. So we need to remove them as well.
* The consolidated dashboard is still slightly broken, for example the SLO table is still messed up. It's also based on an old state of the dashboards. So we need to use the current state of the in-cluster dashboards and use their exact source as base.

I'd suggest the following approach:
* First add the `cluster_id` variable to the existing dashboards. Maybe a `cluster_id` version of the Central dashboard as well that spans all clusters? You didn't do this so far, but would make sense and would be easy.
* Then merge the existing cluster overview + SLO dashboards as you suggested + any other minor modifications.
* Then do the json generation.

So basically I think we should start with the cluster-local dashboards as they are now, and morph them into the consolidated dashboard. Rather then doing it the other way around.

vinaybommana:
Thank you @stehessel I thought I would add the consolidated dashboards initially and remove the old ones if the consolidated dashboard works, I could do the other way around as well

stehessel:
superseded by https://github.com/stackrox/rhacs-observability-resources/pull/131

PR #123, opened 2023-08-03
ROX-18657 Remove CPU limit alert

ludydoo:


resources/mixins/kubernetes/jsonnetfile.lock.json
```diff
@@ -18,7 +18,7 @@
           "subdir": "grafana-builder"
         }
       },
-      "version": "dac760854501e85dd2761160249f08aafff99a85",
+      "version": "c0d0a133a88a1a1b0729525d0269e861a28ce09a",
```
stehessel, line 21: ?

stehessel (approved):

resources/mixins/kubernetes/jsonnetfile.lock.json
ludydoo: Yeah, it does that automatically with `make generate` will revert

stehessel: FYI there is nothing else that `make generate` does, it's just for convenience when updating the mixins. Although maybe we should separate `generate` and `update`...

PR #126, opened 2023-08-04
added rhacs-central.json for in-cluster and cluster-wide central dasboards

vinaybommana:
Hey Team, for adding central dashboard to use rhobs metrics / cluster-wide dashboard, I've added the generated json structure for central dashboard. this will not affect the existing `rhacs-central-dashboard.yaml` grafana dashboard, but both in-cluster and cluster-wide dashboards will be generated from `rhacs-central.json`
Screenshot from 2023-08-04 11-59-07

this will proceed the consolidated dashboard work, I'll be merging the cluster-overview and slo dashboards by creating json and following the same structure

I'm having trouble parsing the `rhacs-central.json` using the command
```
yq ".spec.json = ($(cat generated/dashboards/rhacs-central.json) | to_json)" templates/dashboards/rhacs-central-dashboard.yaml
```
`yq` is throwing me errors that the argument list is too long, since the json is `5226` lines long, I'm not sure how to mitigate this so I've used `sed` and generated the yaml files in a hacky way, please provide any suggestions regarding this

as part of [ROX-18784](https://issues.redhat.com/browse/ROX-18784)


Makefile
```diff
@@ -1,3 +1,4 @@
 .PHONY: generate
 generate:
 	$(MAKE) -C resources/mixins/kubernetes generate
+	$(MAKE) -C resources/grafana generate
```
stehessel, line 4: Hmm this will currently also update the Kubernetes mixin revision if one just wants to generate the dashboard. Although I guess the problem is more with the mixin generate command, we should separate it from the update part. But we can do that in a separate PR.

resources/grafana/generated/dashboards/rhacs-central.json
```diff
@@ -0,0 +1,5229 @@
+{
```
stehessel, line 1: I think the file structure is a bit unfortunate as the `.json` here is the source of truth for the dashbord `.yaml`. So I would rather put the latter into a `/generated` folder and keep the `.json` more top level.

stehessel:
> yq ".spec.json = ($(cat generated/dashboards/rhacs-central.json) | to_json)" templates/dashboards/rhacs-central-dashboard.yaml

This works fine for me. What version of yq are you using?

resources/grafana/rhacs-central-configmap.yaml
```diff
@@ -0,0 +1,5240 @@
+apiVersion: v1
```
stehessel, line 1: Maybe we should put this into a separate directory?

resources/grafana/templates/dashboards/rhacs-central-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
stehessel, line 1: Let's add a comment to these template that explains why there is a `ConfigMap` and a `GrafanaDashboard` version.

vinaybommana:
```
➜  grafana git:(ROX-18784) yq --version
yq (https://github.com/mikefarah/yq/) version v4.34.2
```
I'm facing issues only with very large json files, which in this case `rhacs-central.json` is

resources/grafana/generated/dashboards/rhacs-central-dashboard.yaml
```diff
@@ -1,3 +1,4 @@
+# rhacs-central-dashboard.yaml is generated from rhacs-central.json
```
stehessel, line 1: You have to change the location in the `index.json` as well after moving the file.

stehessel:
> I'm facing issues only with very large json files, which in this case rhacs-central.json is

Strange, I cannot reproduce this with the same version. But I guess it's ok to use sed then.

resources/grafana/generated/dashboards/rhacs-central-dashboard.yaml
vinaybommana: Thank you @stehessel, updated the `index.json`, can you explain the process on how these dashboards will be deployed onto clusters

stehessel: Sure, the observability operator (https://github.com/redhat-developer/observability-operator) has a pull mechanism which pulls the config from a configured git revision.

stehessel (approved):
LGTM, just need to fix the CI linter.

PR #127, opened 2023-08-08
fix(sli): clamp error rate at 1

stehessel:
With the current metrics definition we observed that in rare cases `central:sli:error_rate > 1` for short periods. This can happen if Prometheus reports inconsistent values for incoming traffic, e.g. when 2xx requests are higher than total requests. This is caused by the disparity between processed and in-flight requests. As a result of this bug, the error budget exhaustion may appear smaller than it is.

This PR addresses this issue by:
* refactoring the definition by extracting recording rules. This is helpful when investigating error rates.
* use only gRPC requests that already have terminated without considering in-flight requests.
* adding a `clamp` just in case.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -358,8 +358,8 @@ spec:
 
         - expr: |
             sum by (namespace, rhacs_instance_id, rhacs_org_id, rhacs_org_name, rhacs_cluster_name, rhacs_environment)
-            (rate(grpc_server_started_total{namespace=~"rhacs-.*", job="central", grpc_type="unary",grpc_service!="v1.PingService"}[10m]))
-          record: central:grpc_server_started:total:rate10m
+            (rate(grpc_server_handled_total{namespace=~"rhacs-.*", job="central", grpc_type="unary",grpc_service!="v1.PingService"}[10m]))
+          record: central:grpc_server_handled:total:rate10m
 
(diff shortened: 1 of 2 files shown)


resources/prometheus/prometheus-rules.yaml
```diff
@@ -358,8 +358,8 @@ spec:
 
         - expr: |
             sum by (namespace, rhacs_instance_id, rhacs_org_id, rhacs_org_name, rhacs_cluster_name, rhacs_environment)
-            (rate(grpc_server_started_total{namespace=~"rhacs-.*", job="central", grpc_type="unary",grpc_service!="v1.PingService"}[10m]))
-          record: central:grpc_server_started:total:rate10m
+            (rate(grpc_server_handled_total{namespace=~"rhacs-.*", job="central", grpc_type="unary",grpc_service!="v1.PingService"}[10m]))
```
dhaus67, line 361: Not related to this PR, but it might be worthwhile to consider ignoring the `v1.MetadataService` here as well.

dhaus67 (approved):

resources/prometheus/prometheus-rules.yaml
stehessel: I'm not so sure 😅 . See https://github.com/stackrox/rhacs-observability-resources/pull/134 while ignoring it makes sense, it also creates problems because the traffic to many Centrals is so sparse.

PR #130, opened 2023-08-09
fix(makefile): separate `generate` and `update` commands

stehessel:
Separate `generate` and `update` from each other, such that one can regenerate the mixins without also having to update their revision.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
 .PHONY: generate
 generate:
 	$(MAKE) -C resources/mixins/kubernetes generate
+
+.PHONY: update
+update:
+	$(MAKE) -C resources/mixins/kubernetes update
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,3 +1,7 @@
 .PHONY: generate
 generate:
 	$(MAKE) -C resources/mixins/kubernetes generate
+
+.PHONY: update
+update:
+	$(MAKE) -C resources/mixins/kubernetes update
```
janisz, line 7: Who will call it? How about adding a workflow that will handle that weekly?

stehessel: We don't really need to update the mixins very frequently, only if we want a new feature of a newer version. So I think keeping it at manual casual updates is good enough.

PR #131, opened 2023-08-10
added consolidated dashboards - combining cluster overview and central metrics

vinaybommana:
As part of [ROX-18785](https://issues.redhat.com/browse/ROX-18785) adding consolidated dashboard, which is a combination of cluster slos and central metrics


stehessel:
I think we need to remove the old dashboards to avoid drift. Can you make screenshots to proof that the consolidated is equivalent to the old dashboards?

stehessel:
@vinaybommana Have you made any progress on the PR?

stehessel:
closing as stale

PR #134, opened 2023-08-15
fix(metrics): consider PingService in SLI

stehessel:
Add internal requests via the PingService to the request count considered in error rate SLI. For reference, the PingService adds a baseline of 1 request every 10 seconds.

* (+) SLI are more stable because of a higher base of requests. Especially relevant when Central is essentially idling, because then a small number of failed requests may trigger the SLI. See https://redhat-internal.slack.com/archives/C05LSCJCRMY for an example of this.
* (-) This is also masking real failures with non-relevant internal requests.


stehessel:
Closing for now, not sure if it's the best way forward.

PR #138, opened 2023-08-29
fix(dashboards): only count secured clusters with cores > 0

stehessel:
The current query may count secured clusters that already have disconnected, but are still present in the Central metrics with `0` secured cores. This happens because Prometheus keeps the metrics in memory until Central is restarted.

### DIFF
--- a/resources/grafana/rhacs-cluster-overview-dashboard.yaml
+++ b/resources/grafana/rhacs-cluster-overview-dashboard.yaml
@@ -35,7 +35,7 @@ spec:
       "editable": true,
       "fiscalYearStartMonth": 0,
       "graphTooltip": 0,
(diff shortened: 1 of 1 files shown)


resources/grafana/rhacs-cluster-overview-dashboard.yaml
janisz, line 1: is there a way to keep json outside of this file? Having json in yaml is not my favourite way of storing data :)

stehessel: Yes refactoring is planned anyway as part of https://github.com/stackrox/rhacs-observability-resources/pull/131. For Central dashboard the json is already separated.

PR #141, opened 2023-09-14
Add legend panel in DB metrics row

rukletsov:
This PR restructures widgets in the DB metrics row and adds legend panel for the "Postgres object count" bootstrapped with explanation for the pattern of `network_flows_v2`.

Screenshot 2023-09-14 at 14 00 30

JSON's version and schema version are bumped, other seemingly unrelated changes are manually removed from the exported JSON before the PR has been prepared.

### DIFF
--- a/resources/grafana/rhacs-central.json
+++ b/resources/grafana/rhacs-central.json
@@ -3749,7 +3749,6 @@
             "type": "prometheus",
             "uid": "PBFA97CFB590B2093"
           },
-          "description": "Number of objects in each table",
           "fieldConfig": {
             "defaults": {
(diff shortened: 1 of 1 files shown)


stehessel (approved):
> JSON's version and schema version are bumped, other seemingly unrelated changes are manually removed from the exported JSON before the PR has been prepared.

In my opinion `version` is pretty meaningless for us, because the dashboard is provisioned via our code repository and not via Grafana's internal database. So the Grafana dashboard version does not actually track anything, and cannot be used to e.g. switch between revisions in Grafana.

PR #142, opened 2023-09-18
ROX-19726: Sync jobs for stage and prod.

porridge:
See ticket.


stehessel:
Why was https://issues.redhat.com/browse/ROX-19725 done in the first place? The tickets themselves don't have any context. I'm just wondering what the gain is since I'm not aware of any problems with the current setup, and this change significantly increases the amount of syncing effort. Would the best practice to test obs changes now be to run the sync manually after every merge?

We also immediately deploy fleet-* changes to stage, why not obs changes?

porridge:
> Why was https://issues.redhat.com/browse/ROX-19725 done in the first place? The tickets themselves don't have any context.

Note their parent task, https://issues.redhat.com/browse/ROX-16093, which is a part of the pre-prod env epic.

> We also immediately deploy fleet-* changes to stage, why not obs changes?

Not any more. See https://redhat-internal.slack.com/archives/C0313JYKH8W/p1695030955968949

rukletsov:
> Can you add a note in the README that explains the sync sequence to get changes into prod?

I think best would be to rename actions appropriately (not everyone reads docs). How about:

* `sync-stage-from-master`
* `sync-prod-from-stage`

Hopefully this more or less communicates the sync sequence.

stehessel:
Took the liberty to update the README myself. @rukletsov can you review?

rukletsov (approved):

porridge:
Thanks for taking care of this! Following up on this PR was literally the next item on my TODO list :sweat_smile:

Now, should we add steps to actuate these actions as part of the [new release checklist template](https://docs.google.com/spreadsheets/d/1xh0a8UOAG7skOTgnjRBkX4Y_y2s4Ut2VkthJuXk-HJg/edit#gid=49364065)?

stehessel:
I personally don't like adding even more scope to the acs-cs release. So I'd say separate releases are fine. Imo we also don't need scheduled releases just for dashboard/alerting changes. Just deploy to prod after changes have been observed to work on stage.

PR #146, opened 2023-09-28
no federation of central platform metrics

stehessel:
Central integrates now with OpenShift platform prometheus. As such, we need to exclude these metrics from the platform metrics federation to prevent double counting.


connorgorman (approved):
Seems reasonable to me

PR #149, opened 2023-10-02
fix(sli): allow `Unknown` status code in SLI error rate

stehessel:
This is a workaround to the SLO issue observed in https://issues.redhat.com/browse/ROX-19887.
In short, due to flawed ordering of gRPC interceptors in Central, client errors were counted as `Unknown` gRPC codes in Prometheus metrics. To protect our SLO metrics from ballooning out of control, we temporarily not count `Unknown` errors as server errors.

This is to be reverted again once https://github.com/stackrox/stackrox/pull/7942 is released and deployed to production. See https://issues.redhat.com/browse/ROX-19917 for the task to do exactly that.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -351,9 +351,10 @@ spec:
         # The error rate over the last 10 minutes must be smaller than 35% to count as available.
 
         # GRPC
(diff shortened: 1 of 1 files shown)


0x656b694d (approved):
Thanks!

0x656b694d (changes requested):
Second thought: only Metadata errors should be silenced.

0x656b694d (approved):

PR #162, opened 2023-11-13
ROX-16643: Set acscs-worker role

ludydoo:
We are going to add a worker pool for "infrastructure" workloads on prod-dp-01. 
This node pool will be labeled `node-role.kubernetes.io/acscs-infra`. This will add a role to the node. 
For other clusters, the "default" node pool will be labeled with this as well. 

This will have the consequence that our strictly-worker-nodes query will now return 0 nodes for all clusters apart from prod-dp-01. 

This will add a specific label for those nodes which are considered as "worker" nodes.


stehessel (changes requested):
need to remove the change for local testing again

ludydoo:
@stehessel I added some changes to the `strictly_worker_nodes`. Because in some clusters, the worker nodes will be the same as the acscs-infra nodes, I will add a label to the "regular" worker pools `node-role.kubernetes.io/acscs-worker`. Adapted the queries accordingly.

stehessel (approved):

PR #165, opened 2023-11-13
fix: regenerate dashboards

stehessel:
Regenerate the dashboards after the metrics rename in #162. This fixes the missing widget data in the dashboards.

The changes have been generated by `make generate` on the current master state. I guess we should add this to the pre-commit hook to prevent this in the future.


janisz (approved):

janisz:
>  I guess we should add this to the pre-commit hook to prevent this in the future.

That will be a nice follow up

stehessel:
> > I guess we should add this to the pre-commit hook to prevent this in the future.
> 
> That will be a nice follow up

See #166.

PR #170, opened 2023-11-30
perf: optimize overview dashboard queries

stehessel:
Optimize a few of the Grafana queries in the cluster overview dashboard:

* Only show top 5/10 values dashboards that filter by namespace. This reduces the number of time series that are queried, and also makes the widgets less cluttered. Note that the actual number of different namespaces may still exceed 5/10 if the top namespaces changed during the dashboard time window.
* Filter out useless data in the tenant table.
* Enable pagination for the scanner table.

Screenshot 2023-11-30 at 16 03 26

### DIFF
--- a/resources/grafana/generated/dashboards/rhacs-cluster-overview-configmap.yaml
+++ b/resources/grafana/generated/dashboards/rhacs-cluster-overview-configmap.yaml
@@ -11,2087 +11,2136 @@ data:
   json: |
     {
       "annotations": {
-    	"list": [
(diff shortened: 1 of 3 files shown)


janisz (commented):
Why indentation has changed?

stehessel:
There was a mix between tabs and spaces before. My guess is that somebody edited the `.json` by hand and used an editor which did that. Now it's just the `.json` generated by Grafana, which is all spaces.

PR #172, opened 2023-12-01
ROX-21046: Alerts for tenant nearing OOM

ludydoo:
Adding alerts for tenant containers that are about to oom

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -159,6 +159,38 @@ spec:
             description: "Fleetshard synchronizer manages `{{ $value }}` centrals. The number of Centrals should always be larger than zero in a working system. If it drops to or below zero, fleetshard synchronizer is assumed to be in a failed state."
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-007-fleetshard-sync-reconciliation-error.md"
 
+    - name: tenant-resources
+      rules:
+        - expr: |
+            sum(container_memory_max_usage_bytes{namespace=~"rhacs-.{20}",container!="POD",container!=""}) by (namespace, container, pod)
+          record: rhacs_tenants:namespace:pod:container:max_memory_usage_bytes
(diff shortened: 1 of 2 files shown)


stehessel:
Have you checked how close we are too hitting these alerts on prod right now? I've observed high memory usage on scanner pods in particular. See also https://redhat-internal.slack.com/archives/C0313JYKH8W/p1700683356311839?thread_ts=1700679380.899679&cid=C0313JYKH8W. I'm afraid this will be a very busy alert as is.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -159,6 +159,38 @@ spec:
             description: "Fleetshard synchronizer manages `{{ $value }}` centrals. The number of Centrals should always be larger than zero in a working system. If it drops to or below zero, fleetshard synchronizer is assumed to be in a failed state."
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-007-fleetshard-sync-reconciliation-error.md"
 
+    - name: tenant-resources
+      rules:
+        - expr: |
+            sum(container_memory_max_usage_bytes{namespace=~"rhacs-.{20}",container!="POD",container!=""}) by (namespace, container, pod)
```
stehessel, line 165:
```suggestion
            sum(container_memory_working_set_bytes{namespace=~"rhacs-.{20}",container!="POD",container!=""}) by (namespace, container, pod)
```

ludydoo: Fixed. I've checked if we are hitting the alerts, and we are only for 1 central instance

stehessel: But that means the alert will fire immediately? I think we need to make sure this alert is not noisy, so we need a memory buffer on all Centrals.

ludydoo: I added resources to the 1 central that would've triggered the alert

stehessel (approved):
Please keep an eye and this alert and tighten it if it gets noisy.

PR #174, opened 2023-12-08
ROX-20854: Reduce federated metrics

mtodor:
These changes are trying to solve the problem with large resource consumption (memory) for ACSCS observability Prometheus. The goal is to reduce the number of federated (imported) metrics from OSD monitoring Prometheus.

This PR introduces a script that generates the list of required metrics. It gathers metrics used in dashboards and Prometheus rules and creates the list of required federated metrics.

Changes made:
- added a new script that generated federated metrics YAML file
- added make option that runs only the generation of federated metrics and included this step in the `generate` step
- updated README to improve instructions required for development

**Note**

Reducing federated metrics should also reduce the number of metrics propagated to the RH Observatorium.

**Tests**

1. **EDIT:** ~~The list of required metrics for the consolidated dashboard has been validated, and all metrics are included~~ The consolidated dashboard is not used or maintained. We don't have requirements to federate metrics listed in the consolidated dashboard.
2. Default pre-commit checks in PR


scripts/generate-federate-match.sh
```diff
@@ -0,0 +1,91 @@
+#!/usr/bin/env bash
+
+set -eou pipefail
+shopt -s inherit_errexit
+
+! [ -x "$(command -v jq)" ] && echo 'jq not installed, the script requires it.' && exit 1
```
porridge, line 6: Nit: might be nicer to echo all missing requirements in one go, so they can be installed in one go.
Or just point to the README? 🤔

scripts/generate-federate-match.sh
```diff
@@ -0,0 +1,91 @@
...
+! [ -x "$(command -v mimirtool)" ] && echo 'mimirtool not installed, the script requires it.' && exit 1
+! [ -x "$(command -v realpath)" ] && echo 'realpath not installed, the script requires it.' && exit 1
+! [ -x "$(command -v sort)" ] && echo 'sort not installed, the script requires it.' && exit 1
+! [ -x "$(command -v uniq)" ] && echo 'uniq not installed, the script requires it.' && exit 1
+! [ -x "$(command -v yq)" ] && echo 'yq not installed, the script requires it.' && exit 1
+
+function log() {
+    echo "${1:-}" >&2
```
porridge, line 14: Nit: this way subsequent args will not be silently ignored:
```suggestion
    echo "$@" >&2
```

scripts/generate-federate-match.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    local json_file
+    json_file=$(mktemp)
+
+    yq '.spec' "${os_prom_rule_file}" > "${rules_file}"
+    mimirtool analyze rule-file "${rules_file}" --output="${json_file}"
+    jq '.ruleGroups[].metrics | select( . != null ) | .[]' "${json_file}" --raw-output >> "${metrics_list_file}"
+
+    rm -rf "${rules_file}"
+    rm -rf "${json_file}"
```
porridge, line 41:
```suggestion
    rm -f "${rules_file}"
    rm -f "${json_file}"
```

scripts/generate-federate-match.sh
```diff
@@ -0,0 +1,91 @@
...
+    jq '.dashboards[].metrics[]' "${working_tmp_dir}/acs.json" --raw-output >> "${metrics_list_file}"
+
+    mimirtool analyze dashboard "${repo_dir}"/resources/mixins/kubernetes/generated/dashboards/* --output="${working_tmp_dir}/mixins.json"
+    jq '.dashboards[].metrics[]' "${working_tmp_dir}/mixins.json" --raw-output >> "${metrics_list_file}"
+
+    # Get metrics used in recording rules and alerts
+    get_rules_metrics "${repo_dir}/resources/prometheus/billing-rules.yaml" "${metrics_list_file}"
+    get_rules_metrics "${repo_dir}/resources/prometheus/kubernetes-mixin-alerts.yaml" "${metrics_list_file}"
+    get_rules_metrics "${repo_dir}/resources/prometheus/prometheus-rules.yaml" "${metrics_list_file}"
+    get_rules_metrics "${repo_dir}/resources/prometheus/rhacs-recording-rules.yaml" "${metrics_list_file}"
```
porridge, line 75: Is there a way to discover these dashboard directories and rules files dynamically? I'm worried this list will diverge from reality and cause some frustration due to unexplicably missing metrics...

scripts/generate-federate-match.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    # Filter metrics (exclude metrics that are collected by observability Prometheus or created by recording rules)
+    sort "${metrics_list_file}" | uniq | grep -v -E "^acs|^rox|^aws|^central:|acscs_worker_nodes" > "${metrics_list_file}.filter"
```
porridge, line 78: I don't have any experience in this area, but how can we make sure that the assertion in the comment always stands for example for metrics whose name starts with `aws`?

scripts/generate-federate-match.sh
```diff
@@ -0,0 +1,91 @@
...
+
+    rm -rf "${rules_file}"
+    rm -rf "${json_file}"
+}
+
+function append_consolidated_dashboard_fields() {
+    local federation_config_file="${1:-}"
+    [[ "${federation_config_file}" = "" ]] && log_exit "Variable 'federation_config_file' is empty."
+
+    yq --inplace '."match[]" += [ "container_memory_usage_bytes" ]' "${federation_config_file}"
```
porridge, line 48: What is the connection between `container_memory_usage_bytes` and "consolidated dashboard"? How can we make sure this stays up-to-date going forward?

README.md
```diff
@@ -48,6 +55,13 @@ Then:
 * Update `resources/mixins/kubernetes/mixin.libsonnet`.
 * Run `make generate` to generate the corresponding mixin resources.
 
+### Federated metrics
+
+If you make changes to Alerts or Recording rules and if they include metrics collected by OSD Prometheus, ensure that the federation config includes new metrics.
```
porridge, line 60:
```suggestion
If you make changes to Alerts or Recording rules or dashboards and if they include metrics collected by OSD Prometheus, ensure that the federation config includes new metrics.
```
:point_up: this is also important, right?

porridge (commented):
This seems like a nice trick, but the script seems to assume certain things about metrics flow that might change at any point in the future?

resources/prometheus/federation-config.yaml
```diff
@@ -1,4 +1,172 @@
 match[]:
-  # Federate all platform metrics except ACS metrics. We want all base metrics for the Kubernetes mixin.
-  - '{__name__=~".*", job!="central"}'
-  - '{__name__=~".*", job="central", endpoint!="monitoring-tls"}'
```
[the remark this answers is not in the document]
stehessel: FYI we need to exclude `monitoring-tls` to avoid double scraping metrics from OpenShift monitoring Prometheus (which also scrapes Central).

scripts/generate-federate-match.sh
mtodor: I'm not sure if we can ensure with any "static" checks that some specific metrics are scraped by `rhacs-observability` Prometheus. We know what target we are scraping that is defined in `pod_monitors` (an example: `resources/prometheus/pod_monitors/rhacs-cloudwatch-exporter.yaml`) - but what metric names will be returned by that endpoint, we don't know.

One option is to leave metrics in the list of federated metrics. If they exist in OSD Prometheus, they will be federated; if not, they should not make any impact.

scripts/generate-federate-match.sh
mtodor: I have removed this part.

I talked with Alex and Stephan - the consolidated dashboard is never finished. It is not used or maintained. I'll probably go forward and see to remove it.

resources/prometheus/federation-config.yaml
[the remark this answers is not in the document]
mtodor: This is a really important point! Thanks for noticing it.

I have added a solution to use a filter like this: `{job!~"central|scanner"}`. If we face, in the future, jobs that have these names and metrics required for Alerts or Dashboards, we need to adjust federated metrics.

We could go with a solution to duplicate each metric with both filters:
```
abc{__name__=~".*", job!="central"}
abc{__name__=~".*", job="central", endpoint!="monitoring-tls"}
```
but I think it is unnecessary.

If we would ignore the filter, we could face duplicates because some metrics like `go_memstats_alloc_bytes` - are also returned by the central `/metrics` endpoint.

We could also consider removing job (scrape) configuration for centrals in `rhacs-observability` Prometheus and only rely on federated metrics. But that should be a separate task.

stehessel (commented):
Overall it looks good to me. One concern I have is that when creating dashboards, a common approach is to edit an existing dashboard on the fly. This approach would no longer work for new metrics - one would first have to create/update the dashboard "dry", and then sync the required metrics based on the new dashboard source. Perhaps you can add to way to manually whitelist metrics to make updating dashboards with new metrics a bit more convenient?

resources/prometheus/federation-config.yaml
```diff
@@ -1,4 +1,171 @@
 match[]:
```
stehessel, line 1: Can you mark this file as generated? Maybe also move it to a `generated` folder.

mtodor: I have marked the file as generated. There are also two other generated files in the same folder. If we want to move them to the `generated` subfolder, I would do it in a separate PR.

mtodor:
@stehessel  I have added an option to "manually" define the list of federated metrics. We should use that option sparsely to avoid having everything there, and then automatic generation from Dashboards loses its purpose.

We should add a comment for every metric in the base list on why it's defined there and not automatically generated.

stehessel (approved):

resources/prometheus/federation-config.yaml
```diff
@@ -1,4 +1,171 @@
...
-  # Federate all platform metrics except ACS metrics. We want all base metrics for the Kubernetes mixin.
-  - '{__name__=~".*", job!="central"}'
-  - '{__name__=~".*", job="central", endpoint!="monitoring-tls"}'
+  - :node_memory_MemAvailable_bytes:sum{job!~"central|scanner"}
```
porridge, line 2: I guess we need to keep in mind that once scanner-v4 components land in ACS-CS, we use label `job=scanner` for them 🤔 
cc @mclasmeier

porridge (approved):

resources/prometheus/federation-config.yaml
mtodor: @porridge, it is not relevant to use `scanner` as the job name as long we are not defining that job for Openshift Prometheus. Currently, there are no scanner scrape jobs in OSD Prometheus.

PR #177, opened 2023-12-12
chore: make SLI graphs more readable

stehessel:
Make the SLI graphs more readable by showing a proxy that renders better in Grafana. The problem right now is that Grafana swallows up spikes and instead renders a flat line. The proxy metric works by extending the "downtime" interval via `changes(...[1h])`, such that Grafana has an easier way rendering it. The downside is that downtimes may appear longer than they were - however, the overall percentages (9' notation) are still exact.

Before:
Screenshot 2023-12-12 at 21 07 02

After:
Screenshot 2023-12-12 at 21 06 55

The dashboard can be viewed temporarily under https://grafana-route-rhacs-observability.apps.acs-prod-dp-01.pnz3.p1.openshiftapps.com/d/a3d89363-95c2-470f-82d8-cd9c62fe3840/rhacs-dataplane-central-slos-copy?orgId=1.

### DIFF
--- a/resources/grafana/generated/dashboards/rhacs-central-slo-configmap.yaml
+++ b/resources/grafana/generated/dashboards/rhacs-central-slo-configmap.yaml
@@ -35,7 +35,7 @@ data:
       "editable": true,
       "fiscalYearStartMonth": 0,
       "graphTooltip": 0,
-      "id": 7,
+      "id": 15,
       "links": [],
       "liveNow": false,
       "panels": [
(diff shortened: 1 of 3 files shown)


rukletsov (approved):
I'm stocked every time by the size of the diff a simple change like this generates. This time it is apparently different indentation, are you sure you haven't accidentally messed it up somehow?

Approving the declared change, I resign verifying the diff.

rukletsov (commented):
Ah, it's spaces vs. tabs. Maybe you can fix this one, Stephan.

PR #181, opened 2023-12-14
chore: tighten tenant memory alerts

stehessel:
We have seen that the tenant memory alerts are too noisy in their current state. See https://redhat-internal.slack.com/archives/C0313JYKH8W/p1702483155046459 for a discussion.

Tighten the alerts by:
* constrain to `central` containers.
* increase the thresholds.


kurlov (approved):

GrimmiMeloni (approved):
/lgtm

PR #184, opened 2023-12-19
ROX-21321: Add ACSCS release dashboard

kurlov:
- `make generate` locally also amended kube mixins files
- Add dashboard which might be convenient during canary upgrade of ACSCS
Screenshot 2023-12-19 at 17 19 37

### DIFF
--- a/resources/grafana/generated/dashboards/rhacs-central-release-configmap.yaml
+++ b/resources/grafana/generated/dashboards/rhacs-central-release-configmap.yaml
@@ -1,0 +1,355 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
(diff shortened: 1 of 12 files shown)


stehessel:
> make generate locally also amended kube mixins files

This doesn't happen for me or in CI. What versions of `yq`, `jq` and `jsonnet` are you using? I think we might need to pin them to avoid such inconsistencies.

kurlov:
I use:
- yq 4.21.1
- jq 1.6
- jsonnet, jsonnet-deps v0.20.0

stehessel:
Can you try again with `yq >= 4.40`?

kurlov:
@stehessel this should fix it #185

PR #195, opened 2024-02-07
ROX-21739: SES Monitoring

ludydoo:
Adding monitoring for AWS SES

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -296,7 +296,6 @@ spec:
       rules:
         - alert: AWSSESReputationBounceRateTooHigh
           expr: avg_over_time(aws_ses_reputation_bounce_rate_sum[1h]) > 0.05
-          for: 1s
           labels:
             severity: warning
           annotations:
@@ -305,7 +304,6 @@ spec:
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-040-aws-ses-violation.md"
         - alert: AWSSESReputationComplaintRateTooHigh
           expr: avg_over_time(aws_ses_reputation_complaint_rate_sum[1h]) > 0.01
-          for: 1s
           labels:
             severity: warning
           annotations:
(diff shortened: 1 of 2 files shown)


resources/prometheus/prometheus-rules.yaml
```diff
@@ -291,6 +291,40 @@ spec:
             description: |
               Remaining DB instances: {{ with query "acs_fleetshard_central_db_instances_max - acs_fleetshard_central_db_instances_used" }}{{ . | first | value | humanize }}{{ end }}. A quota increase must be requested from AWS.
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-029-increase-aws-rds-limits.md"
+
+    - name: aws-ses
+      rules:
+        - alert: AWSSESReputationBounceRateTooHigh
+          expr: avg_over_time(aws_ses_reputation_bounce_rate_sum[1h]) > 0.05
+          for: 1s
```
stehessel, line 299:
```suggestion
```
Seems redundant when using `avg_over_time(...[1h])`. It'd either choose the average over time without `for`, or the plain metric `aws_ses_reputation_bounce_rate_sum` with a bigger `for` time.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -291,6 +291,40 @@ spec:
...
+        - alert: AWSSESSendRateTooHigh
+          # cloudwatch is setup to collect aws_ses_send_sum in 2 minute intervals (120 seconds)
+          # and the max send rate is 14/s.
+          expr: (max_over_time(aws_ses_send_sum[1h]) / 120) > 12
+          for: 1s
+          labels:
+              severity: warning
+          annotations:
+              summary: 'AWS SES Send rate too high'
+              description: 'The SES send rate is {{ $value }} per second, which is dangerously approaching the maximum limit of 14 per second.'
```
stehessel, line 324: The description is a bit misleading because the value is the max over the last hour, but it reads like the current send rate value.

stehessel (approved):

PR #199, opened 2024-02-13
feat: add `rhacs_expired_at` to centrals overview panel

0x656b694d:
3 commits:
1. JSON indentation;
2. New column;
3. `make generate`.

image

PR #202, opened 2024-02-15
fix: central overview table position

parametalol:
The change introduced in #199 accidentally moved the "Central Overview Table" out of the "Instance Table" group. This change brings the table back.
The difference can be seen on the stage dataplane Grafana instance vs production.


rhybrillou:
Could you provide before/after screenshots in the description ?

janisz (approved):

PR #210, opened 2024-02-20
chore: add auto-merge in sync

stehessel:
Prevent merge conflicts due to squash commits when syncing `master` -> `stage` -> `production`.

### DIFF
--- a/.github/workflows/sync-prod-from-stage.yaml
+++ b/.github/workflows/sync-prod-from-stage.yaml
@@ -10,8 +10,10 @@ jobs:
         uses: actions/checkout@v4
       - name: Opening pull request
         id: pull
-        uses: tretuna/sync-branches@1.4.0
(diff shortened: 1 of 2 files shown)


dhaus67 (approved):

.github/workflows/sync-prod-from-stage.yaml
```diff
@@ -10,8 +10,10 @@ jobs:
         uses: actions/checkout@v4
       - name: Opening pull request
         id: pull
-        uses: tretuna/sync-branches@1.4.0
+        # We need main for PULL_REQUEST_AUTO_MERGE_METHOD, which is currently not yet released.
+        uses: tretuna/sync-branches@main
```
mtodor, line 14: @stehessel Should we pin it to commit to avoid unexpected changes in main to break our actions?

stehessel: See the comment, we need `main` until they release the `PULL_REQUEST_AUTO_MERGE_METHOD` feature.

PR #213, opened 2024-03-04
fix: automerge action

stehessel:
The `PULL_REQUEST_AUTO_MERGE_METHOD` option on the sync action did not seem to work. It did not enable `auto-merge` as expected. Instead, introduce a dedicated automerge action.

### DIFF
--- a/.github/workflows/automerge.yaml
+++ b/.github/workflows/automerge.yaml
@@ -1,6 +1,9 @@
 name: automerge
 on:
   pull_request:
+    branches:
+      - stage
+      - production
     types:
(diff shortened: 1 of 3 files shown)


.github/workflows/automerge.yaml
```diff
@@ -0,0 +1,30 @@
+name: automerge
+on:
+  pull_request:
```
mtodor, line 3: As I understand this action will apply for all branches. Should we limit it for `stage` and `production` branches?

mtodor (approved):

PR #218, opened 2024-04-15
ROX-23673: Add alerts for cluster autoscaler

mtodor:
This PR adds alerts related to cluster autoscaler behavior, and it removes alerts related to previous manual scaling.

Alerts are taken from default Openshift operator alerts: https://github.com/openshift/cluster-autoscaler-operator/blob/287595b0ee37b893c02d51d3a461ba118b90c3dc/pkg/controller/clusterautoscaler/monitoring.go#L153

Additional documentation related to alerts can be found here: https://github.com/openshift/cluster-autoscaler-operator/blob/287595b0ee37b893c02d51d3a461ba118b90c3dc/docs/user/alerts.md

They have different levels.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -680,41 +680,23 @@ spec:
             cpu_resource_limits:acscs_worker_nodes:by_availability_zone:sum
             /
             availability_zone:acscs_worker_nodes:allocatable_cpu
-        - alert: WorkerNodesMemoryQuotaOverCommitWarning
-          expr: avg(availability_zone:acscs_worker_nodes:memory_request_ratio) > 0.85
-          for: 5m
+        - alert: WorkerNodesMemoryQuotaOverCommit
+          expr: avg(availability_zone:acscs_worker_nodes:memory_request_ratio) > 0.99
+          for: 15m
           labels:
             severity: warning
           annotations:
             summary: "There is a risk of over-committing Memory resources on worker nodes."
-            description: "During the last 5 minutes, the average memory request commitment on worker nodes was {{ $value | humanizePercentage }}. This is above the recommended threshold of 85%."
+            description: "During the last 15 minutes, the average memory request commitment on worker nodes was {{ $value | humanizePercentage }}. This could make pods unschedulable."
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-cloud-service/runbooks/-/blob/master/sops/dp-027-cluster-scale-up.md"
-        - alert: WorkerNodesMemoryQuotaOverCommit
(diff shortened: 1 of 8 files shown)


ludydoo (commented):
Not sure about removing the old alerts. Should we keep them, but put them as warnings with super high (99%) thresholds? @stehessel

mtodor:
@stehessel @ludydoo I have revered old alerts and changed them.

I don't have a strong opinion on whether we should keep old alerts or not. I just wanted to reduce the number of maintained alerts and the potential noise they can produce if they fire for the same reason.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -680,41 +680,23 @@ spec:
             cpu_resource_limits:acscs_worker_nodes:by_availability_zone:sum
             /
             availability_zone:acscs_worker_nodes:allocatable_cpu
-        - alert: WorkerNodesMemoryQuotaOverCommitWarning
-          expr: avg(availability_zone:acscs_worker_nodes:memory_request_ratio) > 0.85
-          for: 5m
+        - alert: WorkerNodesMemoryQuotaOverCommit
+          expr: avg(availability_zone:acscs_worker_nodes:memory_request_ratio) > 0.99
```
stehessel, line 684: I'd suggest to keep the old critical value of `0.95` here. Assuming that we only run into that if the autoscaler is malfunctioning.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -680,41 +680,23 @@ spec:
...
-            severity: critical
-          annotations:
-            summary: "There is a high risk of over-committing Memory resources on worker nodes."
-            description: "During the last 5 minutes, the average memory request commitment on worker nodes was {{ $value | humanizePercentage }}. This is above the critical threshold of 95%."
-            sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-cloud-service/runbooks/-/blob/master/sops/dp-027-cluster-scale-up.md"
-        - alert: WorkerNodesCPUQuotaOverCommitWarning
-          expr: avg(availability_zone:acscs_worker_nodes:cpu_request_ratio) > 0.85
-          for: 5m
+        - alert: WorkerNodesCPUQuotaOverCommit
+          expr: avg(availability_zone:acscs_worker_nodes:cpu_request_ratio) > 0.1
```
stehessel, line 693: Why `0.1`? Won't that trigger all the time?

mtodor: That is a mistake (leftover from some testing). I have fixed it and also updated tests to check that alerts don't fire for low values.

stehessel (approved):

ludydoo (approved):

PR #220, opened 2024-04-18
Renaming metrics for SELinux and NetworkPolicy metrics before ingestion

mclasmeier:


stehessel:
The changes look good to me, but I'd like to see a verification that the relabelling works correctly. Have you tested that it works as expected? You can test this by deploying the cloudwatch exporter + prometheus operator (e.g. by using kube-stack Helm chart https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack).

PR #221, opened 2024-04-22
Add graphs for cluster violations to Grafana dashboard

mclasmeier:
image


stehessel:
Could you please save the dashboard under a new name (copy will be temporary) so it's easier to review the dashboard changes?

mclasmeier:
Saved it additionally as `RHACS Dataplane - Cluster Metrics (+ Cluster Violations)`.

resources/grafana/sources/rhacs-cluster-overview.json
```diff
@@ -1825,6 +1825,435 @@
...
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "PBFA97CFB590B2093"
+          },
+          "disableTextWrap": false,
+          "editorMode": "builder",
+          "exemplar": false,
+          "expr": "sum by() (selinux_denials_sample_count{rhacs_cluster_name=\"acs-int-us-01\"})",
```
stehessel, line 1920: We can't use the hardcoded cluster name here since the same dashboard is used across environments.

resources/grafana/sources/rhacs-cluster-overview.json
```diff
@@ -1825,6 +1825,435 @@
...
+      "panels": [],
+      "title": "Cluster Violations",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "PBFA97CFB590B2093"
+      },
+      "description": "Shows occurrence intensity of SELinux AVC denials.",
```
stehessel, line 1847: Why do you show both the SELinux denials twice? Just from looking at the plots, it seems the `SELinux Violations per minute` is sufficient (same for NetworkPolicy).

resources/grafana/sources/rhacs-cluster-overview.json
stehessel, line 1920: I would remove the `by()` if you don't want to segment by labels.

resources/grafana/sources/rhacs-cluster-overview.json
```diff
@@ -1825,6 +1825,435 @@
...
+            "type": "prometheus",
+            "uid": "PBFA97CFB590B2093"
+          },
+          "disableTextWrap": false,
+          "editorMode": "builder",
+          "exemplar": false,
+          "expr": "sum by() (selinux_denials_sample_count{rhacs_cluster_name=\"acs-int-us-01\"})",
+          "format": "time_series",
+          "fullMetaSearch": false,
+          "hide": false,
```
stehessel, line 1923: I think you can remove the legend as there is only a single value in the plot anyway.

stehessel (commented):
In general, could you add some information to the panel on what to do with this data? What is the expected / normal behavior? Do we expect there to always be zero violations, or are some violations acceptable? How could I get more information on what type of violations these are if I want to follow up?

resources/grafana/sources/rhacs-cluster-overview.json
mclasmeier: Good catch, thank you

mclasmeier: Do you have a suggestion for me on how to interpolate the right cluster name?

resources/grafana/sources/rhacs-cluster-overview.json
mclasmeier: The legend allows selecting by region, right?
How would we handle the case that in different environments we might in fact run in different regions? Should I just sum over all regions?

resources/grafana/sources/rhacs-cluster-overview.json
mclasmeier: Shouldn't that rather be summing by cluster name, to accommodate for the fact that there can be multiple regions?

resources/grafana/sources/rhacs-cluster-overview.json
mclasmeier: The one is a violations-per-minute graph, the other one is a heatmap-style visualization. They are clearly related, but they highlight different aspects.

I am a little bit undecided which one I prefer, that's why currently there are both in the PR / the dashboard.

What I find problematic about the graph is that we are -- in my opinion -- not primarily interested in measuring violations per time as we would be for something like "requests per second". As I understand the situation we we primarily want to know if "violations have occurred" or not. In the heatmap every time unit with a positive number of violations is marked with a visible bar (of the same height). But, agreed, it might make a huge difference for criticality if there is just one violation or if there are 100 -- as an example. Hence, the heatmap also colours the bar, depending on the number of occurrences of violations.

The graph on the other hand could, depending on the distribution of violations, hide situations where only a small number of violations have occurred, if there are other situations where a huge number of violations has occurred. But we should be interested in analyzing both situations.

Reminder: My current assumption here is that we will reach a baseline of zero for these violations. We are currently seeing some violations from scanner, but these should be fixed at some point.

That being said, we can of course discuss further offline. Just wanted to provide a brief sketch of my reasoning here. Bottom line being for me: I see value in both visualizations, but if you are saying that we should only keep one of them, I guess I would vote in favor of the heatmap.

resources/grafana/sources/rhacs-cluster-overview.json
stehessel: I would just remove this, a data plane cluster has only metrics for itself anyway.

stehessel: Each region is its own data plane cluster and therefore prometheus server.

resources/grafana/sources/rhacs-cluster-overview.json
stehessel: The dashboard is per data plane cluster, and therefore per region, not per environment.

resources/grafana/sources/rhacs-cluster-overview.json
stehessel: Tbh I don't really understand what benefit the heatmap plot provides. In mind I would use a heatmap to visualize three-dimensional information - basically its purpose is to add an additional "heat" dimension to an otherwise already two-dimensional plot. But in this case we only have two dimensions (time, number of violations) to begin with, so I would prefer a simple time series.

Take for example this plot

Screenshot 2024-04-23 at 22 28 50

At a glance, the only information I'm getting is when violations have occurred. There may be a difference in the color of the bars, but it's imperceptible to my eyes. Compare now to this view, which contains the same information about when violations occurred, but adds information about how many violations occurred at what time. Much more useful in my opinion. Note that the peak in the middle goes totally unnoticed in the first plot.
 
Screenshot 2024-04-23 at 22 31 50

> The graph on the other hand could, depending on the distribution of violations, hide situations where only a small number of violations have occurred, if there are other situations where a huge number of violations has occurred. But we should be interested in analyzing both situations.

In this case you will see a big bump in the time series immediately grabbing your attention (without having to know what color means "a lot of violations" in the heatmap). You will also see a small blimp that you could zoom in, it would only be obscured if the difference is many orders of magnitude. And if this is the main concern, the best plot would still be a time series but with a log scale.

mclasmeier: Discussed offline, I am fine with removing the heatmap visualizations.

resources/grafana/sources/rhacs-cluster-overview.json
mclasmeier: I think that this was about the heatmap, which is removed.

resources/grafana/sources/rhacs-cluster-overview.json
mclasmeier: Same hare, I believe this referred to the heatmap, which has been removed.

resources/grafana/sources/rhacs-cluster-overview.json
mclasmeier: Same, this was about the heatmap.

mclasmeier:
Preview dashboard updated: https://grafana-route-rhacs-observability.apps.acs-int-us-01.isbr.p1.openshiftapps.com/d/de195dda-1237-4285-af32-1fe0ad3d2ff1/57f53fcb-ad58-5fc3-8113-d8a7ec83161a?orgId=1

mclasmeier:
> In general, could you add some information to the panel on what to do with this data? What is the expected / normal behavior? Do we expect there to always be zero violations, or are some violations acceptable? How could I get more information on what type of violations these are if I want to follow up?

Where should I add these information? From a quick search on the dashboards I haven't found panels including such a description.

stehessel:
You can either put a text box next to the plots like so

Screenshot 2024-04-24 at 15 10 33

or you can use the widget description. For longer text I would use the former.

For example, one might wonder how get to the component that generated the violation one sees on the dashboard. You can also just put a link to some other place where this info is available (e.g. a SOP or something like that).

resources/grafana/sources/rhacs-cluster-overview.json
```diff
@@ -1825,6 +1825,280 @@
...
+        "y": 29
+      },
+      "id": 145,
+      "options": {
+        "code": {
+          "language": "plaintext",
+          "showLineNumbers": false,
+          "showMiniMap": false
+        },
+        "content": "### Description\n\nThis graph shows the occurences per minute of **SELinux AVC denials** on the cluster.\nThese violations are logged on the cluster, propagated to CloudWatch, aggregated by a log metric, retrieved by the cloudwatch-exporter and finally scraped by Prometheus.\n\n**Expected: 0 violations.**\n\nA violation means that the cluster node's SELinux policy prevented a process' actions.\nAs an example, a violation could indicate that a process on the cluster tried to access a file which is SELinux-protected.\n\n### Drill-Down\n\nLog into the cluster's AWS account. All CloudWwatch related resources are located in the `us-east-1` region.\nNavigate to `CloudWatch / Log Insights` and use a query similar to the following on the log group named `acs-<cluster name>.audit`:\n\n```\nfields @timestamp, @message, @logStream, @log\n| filter @logStream like /linux-audit/\n| filter @message like /AVC/\n| sort @timestamp desc\n| limit 1000\n```\n",
```
stehessel, line 1967:
```suggestion
        "content": "### Description\n\nThis graph shows the occurences per minute of **SELinux AVC denials** on the cluster.\nThese violations are logged on the cluster, propagated to CloudWatch, aggregated by a log metric, retrieved by the cloudwatch-exporter and finally scraped by Prometheus.\n\n**Expected: 0 violations.**\n\nA violation means that the cluster node's SELinux policy prevented a process' actions.\nAs an example, a violation could indicate that a process on the cluster tried to access a file which is SELinux-protected.\n\n### Drill-Down\n\nLog into the cluster's AWS account. All CloudWatch related resources are located in the `us-east-1` region.\nNavigate to `CloudWatch / Log Insights` and use a query similar to the following on the log group named `acs-<cluster name>.audit`:\n\n```\nfields @timestamp, @message, @logStream, @log\n| filter @logStream like /linux-audit/\n| filter @message like /AVC/\n| sort @timestamp desc\n| limit 1000\n```\n",
```

stehessel (approved):
LGTM thanks 🚀 . For extra credit add a link to CloudWatch :).

mclasmeier:
Final changes -- board is updated: https://grafana-route-rhacs-observability.apps.acs-int-us-01.isbr.p1.openshiftapps.com/d/de195dda-1237-4285-af32-1fe0ad3d2ff1/57f53fcb-ad58-5fc3-8113-d8a7ec83161a?orgId=1&from=now-24h&to=now

Changes:
* Added CloudWatch Log Insight links
* Removed a bogus "axisWidth" setting, which caused the y-axis (specifically its labels) to be mostly invisible for the first graph.

mclasmeier:
@stehessel Given the approval, I will proceed with merging. Besides having included the cloudwatch links I also fixed a minor inconsistency by removing a bogus `axisWidth` setting in one of the graphs. Thanks for the feedback!

PR #240, opened 2024-05-06
ROX-24055: Annotations for cluster upgrades

ludydoo:


stehessel:
Could you please save a temporary copy of the updated dashboard and share the link? That makes reviewing much easier.

ludydoo:
> Could you please save a temporary copy of the updated dashboard and share the link? That makes reviewing much easier.

@stehessel sure, here it is:
https://grafana-route-rhacs-observability.apps.acs-int-us-01.isbr.p1.openshiftapps.com/d/ddkxpgdojslq8f/rhacs-dataplane-cluster-metrics-ludovic-copy?orgId=1

Though there's not enough retention time to go back to an upgrade where annotations would be visible

resources/grafana/sources/rhacs-cluster-overview.json
```diff
@@ -18,6 +18,17 @@
...
         },
         "type": "dashboard"
+      },
+      {
+        "datasource": {
+          "type": "prometheus",
+          "uid": "PBFA97CFB590B2093"
+        },
+        "enable": true,
+        "expr": "count by(cluster) (count by (git_version, cluster) (label_replace(kubernetes_build_info{job!~\"kube-dns|coredns\"}, \"git_version\", \"$1\", \"git_version\", \"(v[0-9]*.[0-9]*).*\"))) > 1",
```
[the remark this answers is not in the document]
ludydoo: This is the same query as the `KubeVersionMismatch` alert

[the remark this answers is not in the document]
stehessel: I see, there it is an artifact from the mixins having more generalized queries. I would just remove the `cluster` here.

stehessel:
I guess it would be nice to see this in action. Have you actually tested it once and confirmed the rendering?

Another thought: The metric tracks changes to the underlying k8s version. Is this actually equivalent to the OCP upgrades? Couldn't there be OCP upgrades that don't change the k8s version?

resources/grafana/sources/rhacs-cluster-overview.json
```diff
@@ -18,6 +18,28 @@
...
+        "name": "Kubernetes Upgrade",
+        "textFormat": "Kubernetes Upgrade"
+      },
+      {
+        "datasource": {
+          "type": "prometheus",
+          "uid": "PBFA97CFB590B2093"
+        },
+        "enable": true,
+        "iconColor": "purple",
```
stehessel, line 39: Maybe different color for this one?

resources/grafana/sources/rhacs-cluster-overview.json
```diff
@@ -18,6 +18,28 @@
...
+        "expr": "count (count by (gitVersion) (openshift_apiserver_build_info)) > 1",
+        "name": "Openshift Upgrade",
+        "textFormat": "Openshift Upgrade"
```
stehessel, line 42: nit: 
```suggestion
        "name": "OpenShift Upgrade",
        "textFormat": "OpenShift Upgrade"
```

stehessel (approved):

resources/grafana/sources/rhacs-cluster-overview.json
ludydoo: done

resources/grafana/sources/rhacs-cluster-overview.json
ludydoo: done

PR #242, opened 2024-05-10
ROX-23558: Add Network Policy alert

mclasmeier:
This is part of https://issues.redhat.com/browse/ROX-22148.

The SOP is ~~currently under review~~ merged.

Currently we don't expect to have a zero-baseline for network policy violations until GA, hence this PR aims to allow the current non-zero behaviour.

The behaviour we currently observe is related to "probe namespaces", which are short-lived and cause -- for a "short" period of time -- a "small" amount of violations. I have decided to implement to two different alerting triggers:

1. One, which triggers directly in case the violations cross a certain threshold (per minute), currently set to 15 violations per minute.
2. One, which triggers if any positive amount of violations has been observed for longer than 10 minutes.

Would be great if we could do the review in parallel to the other currently ongoing reviews (plus the ACSCS release scheduled for today).


mclasmeier:
For the record, here's a list of timestamps of triggered Network Policy Violations during a probe. Looking at the graphs, this seems to be a peak of such violations, hence it could give us an idea regarding the violations we can expect during probe runs.
image

ebensh:
This LGTM, but will not approve - Stephan can take a look at the Prometheus configuration and approve then.

stehessel (approved):

PR #265, opened 2024-07-03
Increase alerting threshold for networt policy violations

mclasmeier:
The alert is triggering to often currently.

This PR:
* Removes one trigger entirely.
* Increases the threshold for the second remaining trigger.
* 
Here's a graph showing the amount of violations we currently have to treat as being normal:

image


stehessel:
Wouldn't the spike up to 80 in the "normal" picture still trigger the alert?

GrimmiMeloni (approved):
/lgtm

GrimmiMeloni:
> Wouldn't the spike up to 80 in the "normal" picture still trigger the alert?

Yes, but those are to my understanding the ones we are actually interested in.
The "baseline noise" what is actually giving some trouble, as it causes just to many alerts to trigger.

stehessel (approved):

PR #272, opened 2024-07-17
ROX-25003: Add emailsender alerts

kurlov:
Add essential Email Sender alert.
All alerts are marked as warning on purpose. it's better to soak those alerts as warnings
There is an open MR for Email Sender SOPs: https://gitlab.cee.redhat.com/stackrox/acs-cloud-service/runbooks/-/merge_requests/114

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -203,13 +203,13 @@ spec:
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-047-emailsender-ses-send-error.md"
         - alert: RHACSEmailsenderThrottledSend
           expr: |
-            acs_emailsender_throttled_send_email_total > 0
-          for: 5m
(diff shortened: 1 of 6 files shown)


resources/prometheus/prometheus-rules.yaml
```diff
@@ -159,6 +159,59 @@ spec:
...
+          for: 5m
+          labels:
+            severity: warning
+          annotations:
+            summary: "Email Sender container failing sending emails"
+            description: "Email Sender has a send email error rate of {{ $value | humanizePercentage }} over the last 10 minutes."
+            sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-047-emailsender-ses-send-error.md"
+        - alert: RHACSEmailsenderThrottledSend
+          expr: |
+            acs_emailsender_throttled_send_email_total > 0
```
stehessel, line 206: Is there a rate missing here? If this is a counter, it will always fire otherwise.

stehessel (approved):

PR #279, opened 2024-09-05
ROX-21530: Certificate Expiry Dashboard

aaa5kameric:


stehessel (changes requested):
Thanks for providing the demo dashboard. A few observations:

* The table loads way too slow. I believe the reason for this is that the metric `acs_fleetshard_certificate_expiration_timestamp` itself is flawed - it seems to retain all certificates even after their namespaces have been deleted. Combined with the ephemeral instances created by the probe, this creates a very large number of time series. I'd suggest to fix this in fleet-shard sync first - namely stop reporting metrics for deleted tenants.
* The table renders a bit plain for my taste. Some suggestions are column filters, total count as a footer, centered texts, a larger table that spans the entire page width, a new section to separate visually from the instance tables.
* A column that shows `now() - timestamp` would be nice as well. That way it's easier to understand how much time is left until the expiration is reached.

PR #290, opened 2024-12-04
ROX-21530: Update Cert Expiry Dashboard

aaa5kameric:
Version Control got updated--> double instances and 'rox_central_info' is causing both to be shown, updated query expression to account for this.

### DIFF
--- a/resources/grafana/sources/rhacs-cluster-overview.json
+++ b/resources/grafana/sources/rhacs-cluster-overview.json
@@ -46,7 +46,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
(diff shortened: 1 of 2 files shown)


resources/grafana/sources/rhacs-cluster-overview.json
```diff
@@ -46,7 +46,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 26,
+  "id": 16,
```
ludydoo, line 49:
```suggestion
  "id": 26,
```

ludydoo (changes requested):

ludydoo (approved):

PR #291, opened 2025-01-21
ROX-27056: ArgoCD monitoring

ludydoo:
Add federation config and alerting for argoCd


resources/prometheus/federation-config.yaml
```diff
@@ -30,6 +30,9 @@
       - apiserver_request:burnrate6h{job!~"central|scanner"}
       - apiserver_request_terminations_total{job!~"central|scanner"}
       - apiserver_request_total{job!~"central|scanner"}
+      - argocd_app_info{job!~"central|scanner"}
+      - argocd_app_sync_total{job!~"central|scanner"}
+      - argocd_notifications_deliveries_total{job!~"central|scanner"}
```
kovayur, line 35: I don't think the filter makes sense, it has to do with the metrics exposed by central/scanner itself.

```suggestion
      - argocd_app_info
      - argocd_app_sync_total
      - argocd_notifications_deliveries_total
```

resources/prometheus/federation-config.yaml
```diff
@@ -30,6 +30,9 @@
       - apiserver_request:burnrate6h{job!~"central|scanner"}
       - apiserver_request_terminations_total{job!~"central|scanner"}
       - apiserver_request_total{job!~"central|scanner"}
+      - argocd_app_info{job!~"central|scanner"}
```
kovayur, line 33: If I'm not mistaken, this file is generated. Check `README.md` and `generate-federate-match.sh` for details.
Make sure your change will not be overridden after the next generation.

resources/prometheus/prometheus-rules.yaml
```diff
@@ -860,3 +860,89 @@ for the cluster autoscaler. Limits can be adjusted by modifying the cluster auto
             description: |
               A cluster node logged at least {{ $value }} Network Policy ACL denial(s) per minute.
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-cloud-service/runbooks/-/blob/master/sops/dp-044-network-policy-violation.md"
+
+    # Copied from https://github.com/adinhodovic/argo-cd-mixin
+    - name: argo-cd
```
kovayur, line 865: Nice!

resources/prometheus/federation-config.yaml
ludydoo: This file is generated :/

PR #292, opened 2025-01-27
ROX-21531: Configure alert rules for healthchecks

kovayur:
Adjust SLI/SLO prometheus rules by incorporating Central health checks done by prometheus exporter.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -539,6 +539,7 @@ spec:
             )
           record: central:success_rate10m
 
+        # probe_success is a metric that is exposed by blackbox-exporter that periodically calls Central endpoints (routes)
         - expr: |
             probe_success{app="rhacs", endpoint="central-ui"}
           record: central:ui_health
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -539,7 +539,10 @@ spec:
             )
           record: central:success_rate10m
 
-        # probe_success is a metric that is exposed by blackbox-exporter that periodically calls Central endpoints (routes)
+        # probe_success is a metric that is exposed by blackbox-exporter that periodically calls Central endpoints (routes).
+        # The difference between blackbox-exporter checks and the pod readiness probe is the following:
+        # 1. The readiness probe checks the pod internally within the cluster, but blackbox-exporter checks also the public endpoint (route)
+        # 2. The readiness probe doesn't check TLS certificates.
         - expr: |
             probe_success{app="rhacs", endpoint="central-ui"}
           record: central:ui_health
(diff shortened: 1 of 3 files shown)


resources/prometheus/prometheus-rules.yaml
```diff
@@ -539,11 +539,24 @@ spec:
             )
           record: central:success_rate10m
 
+        - expr: |
+            probe_success{app="rhacs", endpoint="central-ui"}
+          record: central:ui_health
+
+        - expr: |
+            probe_success{app="rhacs", endpoint="central-data"}
```
stehessel, line 547: What is the difference between central health and central readiness probe?

resources/prometheus/prometheus-rules.yaml
```diff
@@ -552,14 +565,17 @@ spec:
...
             or on (namespace, rhacs_instance_id) central:sli:pod_ready
           record: central:sli:error_rate
 
         # Availability SLI
         # If any of the SLIs is violated, the service counts as unavailable.
         # This is a time series of 0s (down) and 1s (up).
         - expr: |
-            central:sli:pod_ready * on (namespace, rhacs_instance_id) central:sli:error_rate
+            central:sli:pod_ready
+            * on (namespace, rhacs_instance_id) central:health
```
stehessel, line 577: What happens here if `probe_success` is undefined? How likely is that?

resources/prometheus/prometheus-rules.yaml
```diff
@@ -539,11 +539,24 @@ spec:
             )
           record: central:success_rate10m
 
+        - expr: |
+            probe_success{app="rhacs", endpoint="central-ui"}
```
stehessel, line 543: I assume this is not related to the probe service? Could you please add a comment what is probed here?

kovayur: Yes, it's exposed by the blackbox-exporter. I've added a comment.

resources/prometheus/prometheus-rules.yaml
kovayur: If you mean the **pod** readiness probe, then the difference is the following:
1. It checks the pod internally within the cluster, but blackbox-exporter checks also the public endpoint (route)
2. It doesn't check TLS certificates.

If you mean something else, please specify.

resources/prometheus/prometheus-rules.yaml
kovayur: It could be if we misconfigured `ServiceMonitor`, or if blackbox-exporter is down for whatever reason.
I would rate the likelihood of that happening as pretty low...

resources/prometheus/prometheus-rules.yaml
stehessel: Yes that's what I meant. Could you please add this explanation as a comment as well?

resources/prometheus/prometheus-rules.yaml
stehessel: So this means that if blackbox exporter is down, we are blind on other signals like pod ready as well, right? If the probability for this is very low then this may be acceptable, just wanted to point this out.

stehessel (approved):

PR #296, opened 2025-02-06
ROX-16860: Cover undefined availability metrics in Central SLOs

kovayur:
Add 2 alerts that may indicate that Central SLO metrics are undefined:

1. Compare total central count and the number of central pods;
2. Compare total central count and the number of central healthcheck probes.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -177,23 +177,23 @@ spec:
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-007-fleetshard-sync-reconciliation-error.md"
         - alert: RHACSFleetshardSyncCentralCountPodMismatch
           expr: |
-            sum(acs_fleetshard_total_centrals) != count(central:sli:pod_ready)
-          for: 15m
+            sum(acs_fleetshard_ready_centrals) != count(central:sli:pod_ready)
+          for: 45m
           labels:
             severity: warning
           annotations:
-            summary: "Fleetshard synchronizer manages {{ query \"sum(acs_fleetshard_total_centrals)\" | first | value }} centrals while the central pod count is {{ query \"count(central:sli:pod_ready)\" | first | value }}."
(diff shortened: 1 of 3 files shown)


resources/prometheus/prometheus-rules.yaml
```diff
@@ -175,6 +175,26 @@ spec:
             summary: "Fleetshard synchronizer manages `{{ $value }}` centrals."
             description: "Fleetshard synchronizer manages `{{ $value }}` centrals. The number of Centrals should always be larger than zero in a working system. If it drops to or below zero, fleetshard synchronizer is assumed to be in a failed state."
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-007-fleetshard-sync-reconciliation-error.md"
+        - alert: RHACSFleetshardSyncCentralCountPodMismatch
+          expr: |
+            sum(acs_fleetshard_total_centrals) != count(central:sli:pod_ready)
+          for: 15m
```
stehessel, line 181: When does `acs_fleetshard_total_centrals` increase (upon creation request or when instance is ready)? I'm wondering if 15 minutes is sufficient buffer. I can also imagine multiple Central being created with a bit of delay, which could result in even more delay.

stehessel (approved):

resources/prometheus/prometheus-rules.yaml
kovayur: You're right, it just shows the number of centrals received from Fleet Manager, regardless of their readiness. 
Theoretically these numbers may never match if the number of instances is volatile enough.
 
I'm going to introduce a more accurate metric in fleetshard-sync that counts ready centrals.

kovayur:
Will reopen when the new metric is available in fleetshard-sync

resources/prometheus/prometheus-rules.yaml
kovayur: Added a new metric `acs_fleetshard_ready_centrals` and the time interval has been increased to 45m.

stehessel (approved):

PR #313, opened 2025-05-22
ROX-28713: add alerting for secured cluster drops

johannes94:
We don't get alerted when secured cluster count for tenants or a data plane cluster drops. Because of that incidents affecting the connectivity of secured clusters are usually discovered only by manually checking the dashboards.

This PR adds alerting for significant drops in the secured cluster count.

We define 3 alerts:
1. Tenant specific: a warning if a single tenant looses 20% of it's secured clusters at once, only affects tenants with more than 3 secured clusters
2. Data Plane, warning if a data plane looses 25% of secured cores over all tenants for 2h
3. Data Plane, critical alerts if a data plane looses 50% of secured cores over all tenants for 2h

I tested with manual queries, that those alerts would have fired for the incident from May 5th itn-2025-00118 but won't trigger on usual disconnections for example OCP upgrades or ACS version upgrades.


resources/prometheus/prometheus-rules.yaml
```diff
@@ -1013,3 +1013,42 @@ spec:
...
+            rhacs_tenants:namespace:central:rhacs_secured_cluster_drop >= 1
+          for: 1h
+          labels:
+            severity: warning
+          annotations:
+            summary: Tenant '{{ $labels.namespace }}' lost at least 20% of its secured clusters
+            description: Tenant '{{ $labels.namespace }}' lost at least 20% of its secured clusters
+            sop_url: "TODO: SoP URL"
+        - expr: |
+              sum(rox_central_secured_clusters) / max_over_time(sum(rox_central_secured_clusters)[7d:])
```
ebensh, line 1033: I'm not familiar enough with the prometheus syntax so just double checking, why is the max_over_time above [72h] but here [7d:] - does the sum affect the range specification?

ebensh (approved):
LGTM from the rule definitions - not sure how it will behave in corner case scenarios e.g. newly created cluster in region with 1 tenant with 1 secured cluster (for testing), then we remove that tenant when done testing and trigger critical alert. That said, alerts in that cluster should be silenced until load is present, so probably worth keeping the alerting simple and ignoring spurious alert in that corner case.

resources/prometheus/prometheus-rules.yaml
johannes94: No reasons changed it to 7d: for every query. Now its 7d for everything.

johannes94:
@ebensh I'll work on a solution for those edge cases. I think that will hit us in stage/prod quiet frequently as we mostly have 0 secured clusters then 1 if someone tests something and back to 0 after the test was done.

resources/prometheus/prometheus-rules.yaml
johannes94: Actually that's wrong there is a diff and I did not understand it. [7d:] would yield a vector where the query is evaluated on every single timestamp of a sample rox_central_secured_clusters whereas [7d] would only evaluate once for the current point in time.

johannes94:
@ebensh I adjusted the rules to only fire alerts for clusters with an average SC count over 7d of at least 5.

PR #317, opened 2025-07-03
ROX-29371: change error budget alerting for less noise

johannes94:
The error budget alerting is producing a lot of noise in case of incidents, because it fires for 28 days after the incident though no tenants are down anymore.

This PR changes the budget alert to only fire if central availability was 0 in the last hour.

### DIFF
--- a/resources/prometheus/prometheus-rules.yaml
+++ b/resources/prometheus/prometheus-rules.yaml
@@ -684,7 +684,7 @@ spec:
             message: "High availability error budget exhaustion for central. Current exhaustion: {{ $value | humanizePercentage }}."
             sop_url: "https://gitlab.cee.redhat.com/stackrox/acs-managed-service-runbooks/blob/master/sops/dp-018-rhacs-central-slo-alerts.md"
           expr: |
-            central:slo:availability:error_budget_exhaustion >= 0.9 and central:sli:availability >= 0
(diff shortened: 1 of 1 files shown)


ludydoo (commented):
I'm not sure about this change. `central:sli:availability` is defined as 

```
central:sli:pod_ready
            * on (namespace, rhacs_instance_id) central:health
            * on (namespace, rhacs_instance_id) central:sli:error_rate
```

By alerting only when `min_over_time(central:sli:availability[1h]) == 0`, the condition becomes very strict. For example, even with an `central:sli:error_rate` of 99%, it seems it would prevent the alert from firing. Unless I'm interpreting this wrong?

johannes94:
`central:sli:error_rate` `central:sli:pod_ready` `central:sli:health` are all boolean metrics either 1 or 0.
See definition of the `central:sli:error_rate`:
```
        - expr: |
            max by (namespace, rhacs_instance_id) (
                central:success_rate10m > bool 0.65
            )
            or on (namespace, rhacs_instance_id) central:health
            or on (namespace, rhacs_instance_id) central:sli:pod_ready
          record: central:sli:error_rate
```

In my understanding with my change those allerts would fire:
- If the error budget was > percentage 
- And one of those booleans was 0 within the last hour

ludydoo (approved):
Ok, this makes sense!

PR #340, opened 2026-05-15
ROX-34665: Add Scans per Day by Namespace panel to cluster overview

ebensh:
## Summary

- Adds a new "Scans per Day by Namespace" table panel to the `rhacs-cluster-overview` Grafana dashboard
- Panel placed directly after the Scanner Overview Table at full width (w=24)
- Uses `increase(rox_central_scan_duration_count[1d])` with matrix transform to show daily scan counts per namespace as date columns

image

## Details

The panel uses the existing `$_relativeTime` and `$_timeShift` dashboard variables to control the time window. Results are formatted as a pivot table with namespaces as rows and dates (YYYY-MM-DD) as columns.

## Test plan

- [ ] Verify panel renders correctly in Grafana staging
- [ ] Confirm date columns populate with scan count data
- [ ] Check panel positioning relative to Scanner Overview Table and Certificate Expiry section

🔗 [ROX-34665](https://issues.redhat.com/browse/ROX-34665)

[ROX-34665]: https://redhat.atlassian.net/browse/ROX-34665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/resources/grafana/sources/.claude/settings.local.json
+++ b/resources/grafana/sources/.claude/settings.local.json
@@ -1,0 +1,12 @@
+{
+  "permissions": {
+    "allow": [
+      "Read(//home/ebenshet/workspaces/rox-34665-grafana-display-scans-per-day/github.com/stackrox/rhacs-observability-resources/**)",
+      "Bash(jj status *)",
+      "Bash(jj rebase *)",
+      "Bash(rm -f resources/grafana/generated/dashboards/new-panel.yaml resources/grafana/sources/.claude/settings.local.json)",
+      "Bash(rmdir resources/grafana/sources/.claude)",
+      "Bash(jj commit -m 'ROX-34665: Add Scans per Day by Namespace panel to cluster overview dashboard *)"
+    ]
+  }
+}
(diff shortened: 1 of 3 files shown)


kurlov (approved):

GrimmiMeloni:
Could we also chart this to allow spotting spikes quickly?

resources/grafana/sources/.claude/settings.local.json
```diff
@@ -0,0 +1,12 @@
+{
```
kovayur, line 1: Can we add this file to `.gitignore`?

kovayur (changes requested):

kovayur:
Agree with @GrimmiMeloni , a chart would be more representative than a table

ebensh:
> Could we also chart this to allow spotting spikes quickly?

> Agree with @GrimmiMeloni , a chart would be more representative than a table

I originally had a continuous plot of the metric and asked about this but the answer I got was:

> I wanted to answer how many scans does a customer do in a given day. Just to know how their usage patterns look like.

So I converted to a table aligned to midnight for the daily counts. I can of course go back to a simple increase()[1d] for the rolling amount. Do we want the table *and* the plot or just the plot?

GrimmiMeloni:
I think just the plot would be fine. It's ok if it only shows the daily number like in the table. The whole idea was to simply have something to "eyeball" the usage patterns.

kovayur:
How about a stacked bar chart with dates on the X-axis?

PR #347, opened 2026-07-23
fix: increase federate scrape timeout

johannes94:
Increasing the timeout and scrape interval for federation due to: https://redhat.pagerduty.com/incidents/Q1SDD3SR0UAG71

### DIFF
--- a/resources/prometheus/federation-config-base.yaml
+++ b/resources/prometheus/federation-config-base.yaml
@@ -5,8 +5,8 @@
       namespaces:
         names:
           - openshift-monitoring
-  scrape_interval: 120s
-  scrape_timeout: 60s
+  scrape_interval: 170s
+  scrape_timeout: 150s
   metrics_path: /federate
   relabel_configs:
     - action: keep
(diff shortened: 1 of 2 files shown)


resources/prometheus/federation-config-base.yaml
```diff
@@ -5,8 +5,8 @@
       namespaces:
         names:
           - openshift-monitoring
-  scrape_interval: 120s
-  scrape_timeout: 60s
+  scrape_interval: 170s
```
ebensh: Is there a reason to use the "magic" number 170s over a "round" number like 3 minutes? Seeing something like "170s" always makes me think there's some reason for that specific value, but then I'd expect a comment explaining why that value was chosen. (Similar though not exactly the point: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#res-magic)

johannes94: I was not sure about the behavior if the interval is smaller / equal to the timeout. I wanted 150 for the timeout because my local curl against that endpoint took 2m15s during that incident so that 150s/2.5m should be covered. Then I just added a little bit for the interval.

johannes94: As discussed in slack I changed it to be 120s for the interval and 90s for the timeout.

ebensh (approved):
I had no problem with making it longer, I just wanted it to be 180s instead of 170s :) Up to you.

PR #348, opened 2026-08-04
ROX-35852: do not alert on openshift-custom-domains.*

johannes94:
The custom domains operator is crashing every now and then on `acs-prod-dp-02`.

We are not using that operator and are not interested in this alerts. Additionally it is deprecated anyways.

This PR disables alerting for that namespace, to stop the noise in PagerDuty.


GrimmiMeloni (approved):
LGTM

